PR #4, opened 2026-06-26
fix(mcp): harden send_request injection + correctness fixes across the MCP surface

hahwul:
## Summary

Hardening and correctness fixes for the `gori mcp` server, found by **actually
driving the built binary as an MCP client** (scripted JSON-RPC over stdio
against a seeded project DB and a raw-socket echo/short-delivery test server)
and cross-checked with an adversarial multi-lens code review of the MCP module.
Each finding was reproduced against the real binary before fixing, and every
fix ships with a regression spec.

Four focused rounds (one commit each):

### 1. Reject CR/LF/NUL request injection in the structured `send_request` path
The structured path (`method` + `headers` map) framed the request itself but
wrote header names/values and the method **verbatim**. A tool arg carrying
`\r\n` split one header into many or smuggled a whole second request; a bare
`\n` did the same on lenient origins; a space/`\r\n` in `method` forged the
request line (`GET /admin HTTP/1.1`). Now rejected on the structured path
(method, header names/values, request target). The `raw` path stays byte-exact
by contract and the body is sent verbatim with a matching `Content-Length`, so
neither can smuggle.

### 2. Validate finding-tool args; stop reporting phantom updates
- `create_finding` silently coerced an unrecognised `severity` (a typo) to
  `info` while `update_finding` rejected it — now both reject it via a shared
  helper (absent/blank still defaults to `info`).
- `update_finding` returned `{"updated":true}` even on a complete no-op (only
  `id`, or a field whose value wasn't a usable string) — now returns an error
  listing the updatable fields, and rejects an empty title.

### 3. Version negotiation, URL-host injection, `int()` robustness
- `initialize` echoed **any** client `protocolVersion` verbatim (a client
  probing `"1999-01-01"` concluded gori speaks it) — now echoes only a
  supported revision, else advertises `PROTOCOL_VERSION`.
- A URL host carrying CR/LF (`URI.parse` keeps it in the authority) was written
  into the auto-generated `Host` header — the same injection class as round 1
  but via the URL. Now rejected on both structured and raw paths.
- `int()` truncated a fractional float (`id 5.9` → flow 5, disagreeing with the
  string `"5.9"` which was rejected) and raised `OverflowError` on a huge float
  (`limit 1e19`), surfacing as a generic tool error. Now accepts only integral,
  in-range floats; the rest become `nil` so number/string encodings agree and
  oversized limits clamp.

### 4. Flag a replay response whose body the origin cut short
`send_request` reported `truncated:false` for a response the origin cut short
(a `Content-Length` body that EOF'd early, or a chunked body without its `0`
terminator — the chunked case has no `Content-Length` to cross-check). The
completeness bool was already computed in `Body.stream` but discarded. Now
carried via `Body.read_complete` → `Replay::Result#incomplete?` → a top-level
`incomplete:true` in the result (placed top-level so it survives an empty body,
kept distinct from a `body.truncated` display cap). The HTTP/2 engine keeps the
default; END_STREAM tracking there is a separate concern.

## Testing
- `crystal spec` — **464 examples, 0 failures** (39 new MCP/body specs).
- `ameba` clean on all touched files.
- Each fix verified end-to-end against the real `./bin/gori mcp` binary
  (injection attempts rejected while legitimate requests — custom verbs, spaces
  in header values — still pass; `incomplete` flag confirmed against a
  short-delivery server for Content-Length-cut and chunked-cut, and NOT raised
  for complete or close-delimited responses).

## Considered but intentionally not changed
- Unknown-tool → `isError` (rather than a JSON-RPC error): an existing spec
  documents this as deliberate, so left as-is.
- Top-level JSON array rejected: MCP `2025-06-18` removed JSON-RPC batching, so
  rejecting it is correct.
- Truncated-multibyte-tail text misclassified as base64: low severity and the
  fix is heuristic/risky; left for a follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/spec/mcp_spec.cr
+++ b/spec/mcp_spec.cr
@@ -347,6 +347,20 @@ describe Gori::MCP::Serialize do
     text.valid_encoding?.should be_true
     text.should contain("A")
   end
+
+  it "flags a replay result as incomplete when the origin cut the body short" do
+    head = "HTTP/1.1 200 OK\r\n\r\n".to_slice
+    r = Gori::Replay::Result.new(head, "hi".to_slice, nil, 1000_i64, incomplete: true)
+    out = JSON.parse(Gori::MCP::Serialize.replay_result_json(r))
+    out["incomplete"].as_bool.should be_true
+  end
+
+  it "omits the incomplete field for a complete replay result" do
+    head = "HTTP/1.1 200 OK\r\n\r\n".to_slice
+    r = Gori::Replay::Result.new(head, "hi".to_slice, nil, 1000_i64)
+    out = JSON.parse(Gori::MCP::Serialize.replay_result_json(r))
+    out["incomplete"]?.should be_nil
+  end
 end
 
 describe Gori::MCP::RequestBuilder do
--- a/spec/mcp_spec.cr
+++ b/spec/mcp_spec.cr
@@ -332,4 +332,52 @@ describe Gori::MCP::RequestBuilder do
     args = JSON.parse(%({"url":"/relative"})).as_h
     expect_raises(Gori::Error) { Gori::MCP::RequestBuilder.build(args) }
   end
+
+  describe "structured-path injection guards" do
+    it "rejects CR/LF in a header value (header injection)" do
+      args = {"url" => JSON::Any.new("http://h.test/"),
+              "headers" => JSON::Any.new({"X-Inj" => JSON::Any.new("a\r\nX-Evil: 1")})}
+      expect_raises(Gori::Error, /header.*X-Inj/) { Gori::MCP::RequestBuilder.build(args) }
+    end
+
+    it "rejects a bare LF in a header value (lenient origins split on LF)" do
+      args = {"url" => JSON::Any.new("http://h.test/"),
+              "headers" => JSON::Any.new({"X-LF" => JSON::Any.new("a\nX-Evil: 1")})}
+      expect_raises(Gori::Error) { Gori::MCP::RequestBuilder.build(args) }
+    end
+
+    it "rejects CR/LF in a header name" do
+      args = {"url" => JSON::Any.new("http://h.test/"),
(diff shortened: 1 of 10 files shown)


hahwul:
## Self-review (`/review`, medium effort: 7 finder angles + verify)

Ran a multi-angle precision review over this PR's diff. 21 candidates → 7 distinct
survivors after dedup + verification. Addressed in `e2378ce`:

**Correctness (medium)**
- **HTTP/2 replay never set `incomplete`** — the round-4 truncation signal only
  covered the h1 engine, so a half-delivered h2 response (END_STREAM never seen:
  GOAWAY / RST_STREAM / drop / MAX_BODY) was reported as complete while the
  byte-identical h1 case was flagged. `read_response` now tracks a genuine
  END_STREAM. ✅
- **Target guard missed SP/TAB** — a space in the URL path forged the request
  line (`GET /a b HTTP/1.1`) though the method half already rejected it. ✅
- **Whitespace-padded header name evaded framing dedup** — `" Content-Length"`
  slipped a second, conflicting length / obs-fold line onto the wire. ✅
  (both unified under `reject_token_breakers`: method/header-name/target/host are
  single tokens; header *values* still allow spaces.)

**Honesty (low)**
- **Invalid `flow_id` silently nulled** → now rejected. ✅
- **Fractional required `id` said "missing"** though supplied → now "invalid 'id'". ✅

**Considered, not changed**
- Out-of-range float `limit` → falls back to the default (vs a huge int clamping
  to max): both absurd inputs, both safe; reworded the comment rather than add
  float-boundary code.
- Severity/status parsed twice: a deliberate trade-off to keep `update_finding`
  under the cyclomatic-complexity bar.
- `read_response` complexity warning is pre-existing on `main` (an inherently
  branchy h2 frame dispatcher the repo's `.ameba.yml` already tolerates).

Full suite: **469 examples, 0 failures.**

PR #21, opened 2026-06-27
Test-and-improve audit: fix proxy desync, h2 replay flow control, fuzzer, QL, intercept, certs (21 commits)

hahwul:
## Summary

A two-round, multi-agent **test-and-improve audit of every gori feature**: build → exhaustively exercise the proxy, TUI, and CLI in isolated environments → fix each real bug or unexpected behavior as its own commit → verify. **21 commits**, full suite green (**675 specs, 0 failures**), each fix carrying a regression spec where unit-testable plus end-to-end verification.

- **Round 1** (17 testing agents across every feature area) surfaced 43 issues → fixed 1 high + all 12 mediums.
- **Round 2** (10 regression + adversarial-hunt agents) confirmed all round-1 fixes hold (**0 regressions**) and caught follow-up bugs in them — including 2 new highs — all fixed.

## Correctness / smuggling (proxy + replay)

- **Interim `1xx` response desync** *(high)* — the proxy treated a `100 Continue` / `103 Early Hints` as the final response, stranding the real one on the keep-alive upstream (served to the *next* request → response smuggling) and breaking `Expect: 100-continue`. Now forwards each 1xx and reads on to the final status (RFC 9110 §15.2); same fix in `gori run replay`.
- **`1xx` carrying a declared body** *(round-2)* — a malformed 1xx with `Content-Length`/`Transfer-Encoding` whose "body" embeds a fake response is now refused, not parsed as the response.
- **Intercept Content-Length** — editing a held body now recomputes `Content-Length` (Burp default), and **adding a body to a bodyless request (e.g. a GET)** *(round-2 high)* now frames it — previously it was dropped and the bytes stranded on the upstream (smuggling).
- **Intercept byte-faithfulness** — opening a held message to *view* and forwarding it no longer rewrites bare `LF`→`CRLF` (a TextArea round-trip); only an actual edit normalizes line endings.

## HTTP/2 / TLS

- **h2 replay flow control** *(round-2 high)* — `Replay::H2Engine` never sent `WINDOW_UPDATE`, so any response over the 65535-byte default window hung 30s then failed. Now credits each DATA frame back on the connection + stream — a 1.29 MB cloudflare.com response replays in ~0.4 s (was a timeout). Affects CLI/TUI/MCP replay.
- **IP-literal MITM certs** — leaf certs for IP hosts now carry an `iPAddress` SAN (was `dNSName`), so curl/browsers accept HTTPS-by-IP interception. Zero-padded octets *(round-2)* fall back to a DNS SAN instead of aborting the handshake.

## Fuzzer

- Unbalanced trailing `§` no longer truncates the request body; auto-mark no longer injects a literal `§` for empty values (`§§` collision).
- `clusterbomb` request-count honours the set-0 fallback (no spurious `--force` prompt for a bounded run).

## QL / findings / CLI

- **`size:`** now filters the displayed request+response **total** (was response-only — filtering by the shown size silently missed); adds `reqsize:` / `respsize:`.
- **`body~` / `header~` regex** now scans **past an embedded NUL** (was truncated by `value_text`).
- **Findings export** hardened against captured-traffic injection — a `` ``` `` in a body can't break the code fence; titles/host collapse to one line.
- Clean errors for: a value-option given no value (all subcommands), an unreadable `gori mcp` db, a non-`http(s)` replay/fuzz `--target` scheme.
- `gori run show` now renders captured **WebSocket frames**; `gori run capture` prints/counts a streaming flow (WS/h2/SSE) **once** (was per-message, tripping `--max` and tearing down live connections).

## TUI

- Help-tab body passes `^P` / `:` through (were dead keys); reveal-whitespace toggle no longer blanks the detail pane; intercept empty-state reflects on/off.

## Deferred (documented, not regressions)

Live keep-alive **SSE** incremental capture, **h2→h1** ALPN downgrade for h1-only origins, and per-file capture-lock keying are architectural changes left as documented follow-ups rather than rushed into a heavily-hardened proxy. See the commit history and `audit-2026-06-27` notes.

## Test plan

`crystal spec` → 675 examples, 0 failures. `crystal tool format --check` clean on the diff. End-to-end smoke across capture / gzip-decode / QL / replay / fuzz / findings-export / MCP all pass.

https://claude.ai/code/session_011qvBRr44cwkf2ymqdUewkz

### DIFF
--- a/spec/findings_export_spec.cr
+++ b/spec/findings_export_spec.cr
@@ -1,0 +1,73 @@
+require "./spec_helper"
+require "../src/gori/findings_export"
+
+private def with_store(&)
+  path = File.tempname("gori-fexport", ".db")
+  store = Gori::Store.open(path)
+  begin
+    yield store
+  ensure
+    store.close
+    File.delete?(path)
+    File.delete?("#{path}-wal")
+    File.delete?("#{path}-shm")
+  end
+end
+
+# Returns the markdown with every fenced code block removed, so anything left is
+# "live" Markdown that a renderer would interpret structurally.
+private def strip_fences(md : String) : String
+  out = [] of String
+  fence : String? = nil
+  md.each_line do |line|
+    stripped = line.strip
+    if f = fence
+      fence = nil if stripped == f # closing fence (bare backticks)
+      next
+    end
+    if stripped.starts_with?("```")
+      fence = stripped.gsub(/[^`]/, "") # the run of backticks that must close it
+      next
+    end
+    out << line
+  end
+  out.join("\n")
+end
+
+describe Gori::Findings::Export do
+  describe ".one_line" do
+    it "collapses control characters so a value stays on one line" do
+      Gori::Findings::Export.one_line("pwn\n## FAKE HEADING").should eq("pwn ## FAKE HEADING")
+      Gori::Findings::Export.one_line("a\r\n\tb").should eq("a b")
+      Gori::Findings::Export.one_line("  clean  ").should eq("clean")
+    end
+  end
+
+  describe ".markdown" do
+    it "keeps an attacker-controlled body (``` + headings) inside its code fence" do
+      with_store do |store|
+        id = store.insert_flow(Gori::Store::CapturedRequest.new(
+          created_at: 1_i64, scheme: "http", host: "h.test", port: 80,
+          method: "POST", target: "/", http_version: "HTTP/1.1",
+          head: "POST / HTTP/1.1\r\nHost: h.test\r\n\r\n".to_slice, body: nil))
+        # Body tries to break out of the ```http fence and inject a heading.
+        evil = "```\n## INJECTED HEADING\n```\nmore".to_slice
+        store.update_response(Gori::Store::CapturedResponse.new(
+          flow_id: id, status: 200, head: "HTTP/1.1 200 OK\r\n\r\n".to_slice,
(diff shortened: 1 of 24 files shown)


hahwul:
## Self-review round (8 finder angles → verify)

Ran a multi-agent review of the diff; 8 findings confirmed, all addressed in 4 follow-up commits:

- **Interim 1xx loop unbounded** (proxy + replay) — a hostile upstream streaming endless body-less 103s could spin the fiber / hang replay. Capped at 64 (Burp/nginx-style).
- **1xx forwarded to HTTP/1.0 clients** — RFC 9110 §15.2 forbids it; now read past but forward only to 1.1.
- **h2 replay `rescue nil` too broad** — masked genuine protocol violations as "incomplete"; narrowed to `IO::Error` so `Gori::Error` surfaces as a failed replay.
- **Intercept CL-sync at the wrong altitude** — moved out of the proxy into the editor (`forward_bytes`, dirty-gated). The proxy is byte-exact (P7) again: an unedited forward — incl. a deliberately CL-mismatched smuggling probe and a HEAD/304/204 response — is no longer rewritten; only an actual edit syncs Content-Length.
- **`ContentLength.boundary`** split at the wrong place for an LF-head with a CRLFCRLF body — now takes the first blank line.
- **`capture_printer` dedup set** grew per-flow on a long capture — now bounded to WS flows (the only ones that re-emit).

Full suite: **679 examples, 0 failures.** Re-verified live: 1xx happy-path, h2 large-response replay (81 KB, ~0.4s).

PR #32, opened 2026-06-27
feat(tui): Convert/Comparer default tabs + Convert sub-tabs + space-on-subtab

hahwul:
## Summary

Three related TUI changes (requested):

1. **Convert + Comparer are now default-visible tabs.** Only `Agent` (a coming-soon placeholder) stays in `Chrome::DEFAULT_HIDDEN`. They were previously palette-only power tools.
2. **Convert gains Replay-style sub-tabs.** Each sub-tab is an independent conversion session (input + chain + output view + custom label). The runner-owned strip appears at ≥2 sessions: `^N` new · `^W` close · `^1-9`/`←→` switch · `r` rename. Open sessions **persist across restart**.
3. **`space` on a sub-tab strip opens the command menu.** To make it meaningful in Convert, adds a dedicated `Verb::Scope::Convert` with 7 menu verbs (new/close/clear/copy/cycle-mode/save/load).

## Implementation notes

- New `ConvertSession` model in `convert_controller.cr`; all per-session state threaded through a `cur` accessor. Custom label lives on `ConvertView#name` (uniform with `ReplayView`/`FuzzerView`, so the shared rename prompt works unchanged).
- Persistence: new `Settings.convert_sessions` (`{input, chain, name}` array). Legacy single `convert_input`/`convert_chain` is read and migrated into one session; the scalar fields are no longer written once `sessions` exists. Untouched installs still write no `convert` block.
- `space` arm added to the shared `handle_subtabs_key`, so the menu now also works on the Replay/Fuzzer strips (Notes' scope has no menu verbs → graceful "no commands" toast via the existing empty-guard).
- Convert's body still consumes every key (`handle_body_key` returns `true`), so the new single-letter mnemonics can't collide with literal text — they're reachable only via the space menu + palette.
- `convert_controller.commit` now also runs on tab-switch (`focus_tab`/`cycle_tab`), matching Replay/Fuzzer, so a strip rename survives a non-quit tab change.

## Tests

- `chrome_tabs_spec`: Convert/Comparer default-visible.
- `space_menu_spec`: Convert scope lists its actions + menu_key uniqueness across all scopes (Convert added).
- `settings_spec`: `convert_sessions` round-trip + legacy-file migration; untouched install omits the block.
- Full suite green (780 examples, 0 failures). Manually smoke-tested in a pty: tab bar, sub-tab create/switch/rename, space menu invoking a verb, persistence across restart.


hahwul:
### Self-review (high-effort, 8 finder angles + verify)

Ran a structured review over the branch diff. The core refactor came back clean (the `cur` accessor is provably never called on an empty `@sessions`; index math in close/jump/move is clamped; the settings round-trip + legacy migration are symmetric; all 3 `ExecContext` subclasses implement the 7 new methods). Fixed the findings worth acting on:

- **`^S` persisted a stale/empty sessions block** → `save_chain` now flushes the live sessions before `Settings.save` (shared `session_tuples` helper). Verified via the space-menu "Save chain" path: the file now contains the in-progress `{hello, base64}` session.
- **Blank-omit invariant narrowed** → a single blank+unnamed session is "nothing to persist" again, so a cleared/dirtied-but-empty workbench writes no `convert` block (added spec).
- **Commit-on-leave gaps** for the now-hideable Convert tab → commit on `settings:tabs` snap-off and `focus_pane` body-leave (mirroring Replay/Fuzzer).
- **Stale comments** corrected (strip is Replay/Fuzzer/Notes/Convert; Comparer no longer "hidden by default").

Deliberately not changed: the `@chain_pre` IME-preedit-on-switch (discarding an in-flight composition on an explicit switch is acceptable), eager `Convert.run` for restored sessions at startup (bounded), and the `space`-needs-≥2-sessions reachability (matches the help text — single-session uses the palette / `^`-chords). Full suite green (780 examples, 0 failures).

PR #33, opened 2026-06-27
feat(tui): first-run setup wizard (bind / theme / AI) + `gori wizard`

hahwul:
## Summary

Adds a **first-run setup wizard** — a standalone, step-by-step TUI that helps a brand-new user configure gori before the project picker, and is re-runnable anytime via the now-wired `gori wizard` command.

It walks through three concerns plus a summary:

1. **NETWORK** — proxy bind IP / port (default `127.0.0.1:8070`; tells the user it's changeable later via `settings:network`)
2. **THEME** — scrollable theme list with a **live preview panel** that renders a mock History view in each theme's own palette, re-theming the whole UI live as you cycle
3. **AI PROVIDER** — the feature doesn't exist yet, so this step is **informational only**: it asks "set up an AI provider?" (yes/no) and shows a "coming soon" note. The answer is **not persisted** (no settings.json change).
4. **REVIEW** — recap + a pointer to `settings:network` / `settings:theme` for later changes.

### Behavior
- **Auto-launches once on first run** — gated on `!File.exists?(Settings.path)`. Both finish and skip call `Settings.save`, so settings.json exists afterward and the wizard never auto-launches again.
- **Re-runnable** via `gori wizard` (replaces the long-standing placeholder), even when settings.json already exists.

### Design notes
- **Standalone component** (`Gori::Tui::SetupWizard`) modeled on `ProjectPicker` — its own run loop, render/sync diffing, Key/Mouse/Preedit/Resize handling, IME-aware `Screen#input_line`. Config-only: no Session / proxy / CA.
- **Staging model** — edits live in wizard-local fields and commit to `Settings` only on finish, so **Esc = skip is coherent**: it reverts the live theme preview to the captured baseline and leaves Settings untouched.
- Auto-launch sits inside `App#run_tui`'s `begin` block so `ensure term.close` restores the terminal if the wizard ever raises. The bind prefill inherits a `gori tui --port` flag.
- `gori wizard` bootstraps its own terminal directly rather than going through `App` (which eagerly loads the CA — heavyweight and unneeded here).

## Files
- `src/gori/tui/setup_wizard.cr` — new `SetupWizard` component
- `src/gori/app.cr` — auto-launch on first run
- `src/gori/cli.cr` — wire up `gori wizard` + help text

## Testing
- `crystal build` clean · `crystal tool format` clean · ameba clean on new code
- **778 specs, 0 failures**
- tmux end-to-end (isolated `GORI_HOME`): first-run auto-launch; all four steps render; theme preview + live re-theming; persistence (`bind_port`, `theme`, no `ai` key); no-relaunch on 2nd start; `gori wizard` manual relaunch; skip reverts theme & doesn't persist it; invalid-port guard; persisted bind takes effect in the runner top bar; too-small-terminal guard.

### DIFF
--- a/src/gori/app.cr
+++ b/src/gori/app.cr
@@ -17,6 +17,7 @@ require "./verbs/comparer"
 require "./tui"
 require "./tui/runner"
 require "./tui/project_picker"
+require "./tui/setup_wizard"
 
 module Gori
   # Top-level orchestrator. Owns the shared cert authority + verb registry and
@@ -47,6 +48,11 @@ module Gori
       term.enable_mouse if Settings.mouse # SGR-1006 click + scroll-wheel nav; one enable covers both the picker and the runner (same term). Runner reconciles live on settings save; term.close disables on exit.
 
       begin
+        # First-run onboarding: no settings.json yet → walk the user through bind /
+        # theme / AI setup once. Inside `begin` so the `ensure term.close` restores
+        # the terminal if it raises. The wizard persists settings.json (even on skip),
+        # so it never auto-launches again.
+        Tui::SetupWizard.new(term).run unless File.exists?(Settings.path)
         loop do
           project = Tui::ProjectPicker.new(term, projects).run
           break unless project # nil => quit gori
--- a/src/gori/cli.cr
+++ b/src/gori/cli.cr
@@ -18,7 +18,8 @@ module Gori
   # - `gori export ca-cert`         → print CA cert path (refactors old --export-ca)
   # - `gori run <sub>`              → non-interactive CLI (see Gori::CLI::Run)
   # - `gori mcp`                    → MCP (Model Context Protocol) server over stdio
-  # - `gori wizard` / `gori update` → placeholders for future work
+  # - `gori wizard`                 → interactive first-run setup wizard (bind/theme/AI)
+  # - `gori update`                 → placeholder for future work
   #
   # Old flat flags (`gori --headless`, `gori --export-ca` ...) continue to work
   # via the tui path for backward compatibility.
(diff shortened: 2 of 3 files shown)


hahwul:
## Review (8 finder angles + verify)

**Fixed:**
- **Short-terminal overflow** (Bind/AI/Review steps): on a short terminal the card height clamps but those fixed-layout steps drew content at fixed offsets, spilling onto the card's bottom border / footer (e.g. at 80×11 the "Default 127.0.0.1:8070…" line overwrote the border). Fixed in 6caa227 with accurate `content_rows` + a `step_fits?` guard that shows the "terminal too small" message instead of a broken card. The theme step is exempt (its list viewport derives from the card height, so it scrolls and fits any usable size).

**Reviewed and intentionally left as-is:**
- *Back-nav keeps the previewed theme* — by design; the staged selection stays applied until skip/finish.
- *IP not validated* — matches existing `SettingsView#save` behavior (only the port is validated; a bad host fails at bind time).
- *`finish` port fallback without range check* — `@port` is re-validated on every forward transition out of the Bind step, so it can't reach Review invalid.
- *`draw_swatch` / scroll-follow duplication with `SettingsView`* — small, intentional (different two-column layout); both are private and may diverge.
- *`Theme.available` allocation per frame* — negligible for a wizard (renders only on input, not continuously).
- *First-run gate via `!File.exists?(Settings.path)`* — accepted tradeoff (a one-shot signal; `gori settings` materializing the file first is a fine edge).
- *`gori wizard` logging* — `Termisu.new` routes `::Log` to a file; verified no STDOUT/alt-screen corruption in the live tmux runs.

Re-verified after the fix: build clean, ameba clean, **778 specs pass**, and tmux runs confirm too-small at h<13–14, full render above, theme step still scrolls at h=12.

PR #55, opened 2026-06-30
feat(cli): add `gori run notes` to read project notes headlessly

hahwul:
## Summary

Adds a `notes` subcommand to the non-interactive `gori run` CLI so the project's Notes-tab documents can be read headlessly (reports, grep, piping into other tools) — the read-only counterpart to the TUI Notes tab.

- `gori run notes` — list one line per note: index, title (first non-blank line), line/byte count; `*` marks the active note
- `gori run notes <n>` — print note `<n>` (1-based) in full
- `gori run notes --all` — print every note in full
- `--format text|json`, `--project NAME` / `--db PATH` (same resolution as the other read subcommands)

Read-only: no DB writes, no capture lock → safe alongside a live capturing instance (SQLite WAL), consistent with `history`/`show`/`findings`.

## Design

- New shared `Gori::Notes` module (`src/gori/notes.cr`) owns the `notes.docs` JSON format (+ legacy single-note fallback), `parse`/`load`/`serialize`/`title`/`line_count`. **Single source of truth**: the TUI `NotesView` now delegates its load/parse/serialize to it instead of duplicating the format.
- Output formatting lives in `CLI::Output` (matching `prism`/`fuzz`/`mine`), making it unit-testable.
- `Run.dispatch` split into a thin EPIPE wrapper + `dispatch_subcommand` to keep the subcommand `case` under the cyclomatic-complexity bar.

## Testing

- `crystal spec` — **964 examples, 0 failures** (incl. 13 new: `Gori::Notes` parse/load/title/line_count + `CLI::Output` note row/json formatting; existing `NotesView` specs still pass after the refactor).
- Ameba clean on edited files (only the pre-existing tolerated `cmd_replay` 13/12 and the conventional `set_preedit` naming remain); `crystal tool format --check` clean.
- End-to-end on a seeded DB: list/show/`--all` in text+json, active-note marker, blank-note title fallback, out-of-range/invalid-index/conflicting-flags/bad-db exit codes (all exit 1), empty project (`no notes` / `[]`, exit 0), and EPIPE-to-`head` (clean exit, no backtrace).

### DIFF
--- a/spec/cli_run_spec.cr
+++ b/spec/cli_run_spec.cr
@@ -242,4 +242,113 @@ describe Gori::CLI::Output do
     Gori::CLI::Output.human_size(5_368_709_120_i64).should eq("5.0GB")     # 5 GiB
     Gori::CLI::Output.human_size(2_199_023_255_552_i64).should eq("2.0TB") # 2 TiB
   end
+
+  it "formats a note listing row: 1-based index, title, '*' for the active note" do
+    row = Gori::CLI::Output.note_row_text(1, "scope\nmore", current: true)
+    row.should contain("* 2")                                                           # 0-based 1 → shown as #2
+    row.should contain("scope")                                                         # title = first non-blank line
+    row.should contain("(2 lines, ")                                                    # plural
+    Gori::CLI::Output.note_row_text(0, "x", current: false).should contain(" 1")        # no '*'
+    Gori::CLI::Output.note_row_text(0, "x", current: false).should contain("(1 line, ") # singular
+  end
+
+  it "falls back to 'note N' in a row for a blank note" do
+    Gori::CLI::Output.note_row_text(2, "   \n\t", current: false).should contain("note 3")
+  end
+
+  it "emits a single-note JSON object with the documented fields (text only when asked)" do
+    full = JSON.parse(Gori::CLI::Output.note_object_json(0, "Title\nbody", current: true, with_text: true))
+    full["index"].as_i.should eq(1)
+    full["title"].as_s.should eq("Title")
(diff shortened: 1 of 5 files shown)


hahwul:
## /review — medium effort (3 correctness + 3 cleanup + altitude + conventions angles, parallel + verified)

**No correctness, removed-behavior, cross-file, STDOUT-purity, or convention bugs found.** Highlights from verification:

- **Behavioral equivalence of the `NotesView` refactor** — all six load paths (valid/malformed JSON, empty `notes` array, legacy key, nothing stored, out-of-range `cur`) reproduce the old behavior exactly; `Notes.serialize` is byte-identical to the old inline `serialize`.
- **No off-by-one** in `cmd_notes` — `parse_note_index` guarantees `n>0` and the `n <= doc.size` guard bounds `idx = n-1` into range.
- **Title / line_count parity** with `TextArea#first_nonblank_line` confirmed (split on `\n`, `\r`-trimmed, blank-aware).
- **dispatch → dispatch_subcommand split** keeps the EPIPE rescue effective and is behavior-preserving.
- No name collisions, no require cycles, clean compile, full suite 964/0.

**Applied review finding (commit ad63ad6):** the PR claimed single-source-of-truth, but `Note#label` (TUI) still derived the title independently of `Notes.title`, so the sub-tab label and the CLI listing could drift on blank/CRLF input. `Note#label` now delegates to `Notes.title`. Behavior-preserving; suite still 964/0.

Informational (not a bug): an out-of-range `cur` in hand-edited `notes.docs` yields no `*` marker in the CLI listing; unreachable in practice since the TUI clamps `cur` before writing.

PR #56, opened 2026-06-30
feat(cli): add `gori run sitemap` headless endpoint tree

hahwul:
## Summary

Adds a `sitemap` subcommand to the non-interactive `gori run` suite — the headless equivalent of the Sitemap TUI tab. It prints the deduplicated **host → path endpoint tree** built from captured flows, with the same shape as the interactive tab (path normalisation, numeric folding, path tags, endpoint counts).

```
$ gori run sitemap
acme.test  (5 paths)
├─ /  [GET]
├─ api  # payment flow
│  ├─ orders  [POST]
│  └─ users  [GET]
│     └─ 123  [GET]
└─ health  [GET]

shop.test  (12 paths)
└─ p
   └─ [1001, 1002, 1003 … +9]  (12 values)
```

## What's new

- **Shared `Gori::Sitemap` module** (`src/gori/sitemap.cr`): the pure tree model + algorithms (`Node`, `build`, `normalize_path`, `stamp_tags!`, `group_sequences!`, `endpoint_count`) **extracted from `Tui::SitemapView`** so the CLI and TUI share one source of truth. The view now delegates to it — behaviour-preserving (all existing sitemap/mouse specs still pass).
- **Formats:** `text` (indented tree, default) · `json` (nested, stable machine contract — folded groups retain their children) · `paths` (flat one-line-per-endpoint, pipe/grep-friendly).
- **Options:** `--project`/`--db`, `-q`/`--query` (positional QL too; un-compilable query rejected like `history`), `-n`/`--limit`, `--in-scope` (host-level scope filter, same logic as the TUI marker), `--no-group` (disable numeric-sequence folding).
- Read-only — opens the store directly, takes **no capture lock**, so it's safe alongside a live capturing instance (SQLite WAL). STDOUT-purity: results → STDOUT, notes/errors → STDERR; clean EPIPE handling.
- Split `Run.dispatch` into a thin EPIPE wrapper + `dispatch_subcommand` to keep both methods under the cyclomatic cap when adding the new `when`.

## Testing

- New `spec/sitemap_spec.cr` (pure builder: build/normalize/group/count/stamp) + `CLI::Output` sitemap text/json/paths cases in `cli_run_spec.cr`.
- **Full suite: 962 examples, 0 failures.** `ameba`/`crystal tool format` clean on touched files (only the pre-existing `cmd_replay` 13/12 remains).
- Verified end-to-end on a seeded DB: text/json/paths, positional + `--query` QL, `--in-scope` (with/without rules), `--limit`, `--no-group`, absolute-form target normalisation, bad QL/db/format, empty results, EPIPE-to-`head`, all exit codes.


hahwul:
## Review summary (medium-effort, 6-angle: correctness ×3, reuse/simplification, efficiency, altitude/conventions)

**No correctness bugs survived verification.** The implementation is solid — full suite **962/0**, ameba/format clean on touched files (only the pre-existing `cmd_replay` 13/12 remains), and end-to-end verified (text/json/paths, positional+`--query` QL, `--in-scope` with/without rules, `--limit`, `--no-group`, absolute-form normalisation, bad QL/db/format, empty results, EPIPE-to-`head`).

Notable candidates that were **refuted**:
- *Bare module-method calls (`add`, `endpoint_count`, …) "might not resolve"* → Crystal resolves bare sibling-method calls; build compiles and `sitemap_spec.cr` asserts the recursion (endpoint_count, fold, nested tags).
- *Unqualified `Sitemap::Node` from `Gori::CLI` might be a different type* → one `Sitemap` (= `Gori::Sitemap`); qualified/unqualified names are the same type; e2e specs mix both.
- *STDERR EPIPE unhandled* → the `dispatch` rescue catches `IO::Error` from any stream and exits 0.
- *`--in-scope`/no-rules could crash* → `Scope#host_in_scope?` does `return false if @rules.empty?`; warned, not crashed.
- Space-vs-comma method joins, `Scope.load` nil-check — intentional / impossible.

**Applied as follow-ups** (commit `cc929c4`, no behavior change):
1. Reworded the `collect_sitemap` comment — it mirrors `SitemapView#reload`'s *order*, but `--in-scope` deliberately uses host-level `Scope#host_in_scope?` (enabled-flag-independent), which is coarser than the TUI lens's per-flow SQL filter. The old wording overclaimed parity.
2. Removed the now-dead `require "uri"` from the view (URI normalisation moved to `Gori::Sitemap`).
3. Documented that JSON keeps a folded group's children nested (complete tree + `grouped:true`), unlike the collapsed text view.

Low-severity style suggestions (single-use helpers, splitting the three format walkers, sharing the QL-EMPTY check) were considered and left as-is — the codebase favours small focused `CLI::Output` helpers, and `cmd_history`'s QL check differs (it aborts mid-open-store; `sitemap_filter` validates before opening).

hahwul:
Rebased onto latest `main` (which gained `gori run prism` #54 and `gori run notes` #55 in the meantime). Resolution:
- `gori run sitemap` slots in alongside `prism`/`notes` (dispatch, help, `parse_format` `paths`, output formatters, specs).
- `main`'s `dispatch_subcommand` was already at the cyclomatic cap (11 `when` arms = 12); adding `sitemap` would hit 13, so the `case` is split into `dispatch_subcommand` + `dispatch_subcommand2` — the same split-to-stay-under-the-bar precedent the repo already used for `dispatch` → `dispatch_subcommand`.

Post-rebase: full suite **981/0**, ameba/format clean on touched files (only the pre-existing `cmd_replay` 13/12 remains), end-to-end re-verified (sitemap text/json/paths coexisting with prism/notes, unknown-subcommand routing through the split).

PR #57, opened 2026-06-30
feat: decode SAML/JWT/GraphQL/form + split-editor replay

hahwul:
## What

Surface encoded protocol payloads as **decoded, editable** views — for inspection in History and for **tampering during security testing** in Replay.

### History — decoded detail panes (mirror the SSE `EVENTS` pane)
New derived panes appear when a flow carries the protocol (no DB table; parsed on demand):

- **SAML** — `SAMLRequest`/`SAMLResponse` from the request body (HTTP-POST), URL query (HTTP-Redirect, raw-DEFLATE), or a response auto-POST HTML form → pretty XML.
- **JWT** — scans `Authorization: Bearer`, `Cookie`/`Set-Cookie`, query, and bodies (`eyJ`-anchored) → decoded header/payload + alg/exp. (Real value over Pretty's whole-body decode, which misses tokens in headers/cookies.)
- **GRAPHQL** — `operationName` + query + variables (POST JSON or GET `?query=`).
- **PARAMS** — `x-www-form-urlencoded` / `multipart/form-data` body + query string, URL-decoded.

### Replay — split editor (the main ask)
The REQUEST column splits into **ENVELOPE** (the full request — headers, target, other params, all editable) and **DECODED** (the payload — SAML XML, or GraphQL query + variables). `^T` toggles the active sub-pane (which enlarges).

On send, the decoded payload — *if edited* — is re-encoded back into the request (SAML param via `replace_param`; GraphQL JSON body via `recompose`, preserving `extensions`) with **Content-Length resynced**, so **envelope edits and payload edits both reach the wire**. Sent as a normal request (ordinary response pane). Session-only, like the WS/gRPC modes.

This replaces the earlier SAML-only "payload hijacks the whole editor" mode, which locked you out of the headers/RelayState/target — and makes GraphQL replay friendly (edit a clean multi-line query instead of an escaped JSON string).

### CLI / MCP
`gori run show` (text + JSON) and MCP `get_flow` emit the same decodes through one shared `DecodedView` emitter; CLI honours `--request-only` / `--response-only`.

## How it's built
- Four **Store-free** decoder modules — `src/gori/{saml,jwt,graphql,form_data}.cr` — operating on primitives (testable, reused by History/Replay/CLI/MCP).
- `Saml.encode_value`/`decode_value`/`replace_param` drive the SAML re-encode; `Graphql.parse_display`/`recompose` round-trip the GraphQL editor.
- A rebindable verb `replay.toggle-decoded` (default `^T`).

## Verification
- **978 specs pass** (new `spec/protocol_decode_spec.cr` + verb-routing); ameba net-neutral on `src/`; `crystal tool format` clean.
- Live tmux + **byte-captured replays**: GraphQL send → envelope preserved, decoded-query edit reached the wire as **valid JSON**, Content-Length exact; SAML send → re-encoded param + preserved `RelayState`, Content-Length exact.
- Adversarial multi-agent review; fixed findings incl. a **crafted-JWT `exp` `Time.unix` crash** (would have taken down the History render / CLI / MCP) and CLI/MCP emitter duplication + divergence.


hahwul:
Follow-up on the split-editor UX (per review):

- **Vertical navigation** — ↓ off the ENVELOPE's last line now crosses into DECODED, ↑ off the DECODED top crosses back (↑ off the ENVELOPE top still pops to the tab bar). `^T` still jumps directly; clicks land in the pane clicked.
- **ENVELOPE ↔ DECODED sync** — the two are kept consistent at every switch boundary (^T / arrow-cross / click / before send·fuzz·mine). The ENVELOPE is the authoritative wire request: leaving DECODED commits the re-encoded payload into it (Content-Length resynced); entering DECODED re-decodes the envelope's current param, so an envelope-side edit shows up decoded. A broken/mid-edit envelope never clobbers DECODED; an untouched payload still sends byte-faithfully.
- Also: `^E` edits the active sub-pane; fuzz/mine flush pending payload edits into the envelope first.

New `spec/tui/replay_decode_spec.cr` covers both sync directions, the send-time commit, byte-faithful pass-through, the SAML re-encode (RelayState + CL), and the arrow-crossing. Full suite 985 green.

PR #74, opened 2026-07-03
feat(tui): keyboard-flow consistency, keymap-drift fixes, roomier space menu

hahwul:
## Summary

A UX pass over gori's keyboard flow — the result of a source audit, implementation, and an adversarial multi-agent review of the diff. Focused on consistency and discoverability, no behavior removed.

### Keyboard-flow consistency
- **`j/k` parity** in every read-only/nav pane that lacked it: Replay response, Fuzzer results+detail, Miner results+detail+summary, Convert output, and the Project scope + host-overrides lists. `j/k` already worked in the browse tabs; now it works everywhere a tester reads.
- **Prism detail keyboard scroll** — previously mouse-wheel only (a keyboard dead-end for long issue detail). New `PrismController#handle_body_key`, gated on `detail_open?` so the list keeps its scoped-verb nav.
- **Comparer `h/l`** pane toggle (it already had `j/k`).
- **Miner sub-tab rename** wired up — it was fully implemented but unreachable dead code.
- **Hint fixes**: Replay response now shows `↵/^R send` (Enter silently re-fired a request), Fuzzer results shows `v dist`, the no-op `^B ws` is dropped from Notes/Project editors, and Help lists the `n` notification center.

### Keymap drift
- The **command palette** chord column now resolves through the effective keymap (`Hotkeys.binding_for`) so a rebind is reflected — filtered the same way the dispatch keymap is, so a hand-edited FIXED-id binding can't advertise a chord that never dispatches.
- **`Registry#validate_menu_keys!`** makes per-scope space-menu key uniqueness a boot-time invariant (was convention-only, guarded only by a spec).

### Space menu sizing
- Raised `SpaceMenu::MAX_ROWS` 12 → 16 (History's Body scope has **13** entries — "Mine parameters" was clipped at 12) and added `▲/▼/↕` scroll markers so clipping is visible on short terminals.

## Notes
- A durable-notification reframe explored during this work was **dropped** in favour of #73's merged approach (which gates the completion toast and the log together by `NotifyMode`). This branch is rebased on top of #73 and does not touch that logic.
- `Help`/status-bar hints are deliberately **not** converted to derive from the keymap — they're free-form prose; the palette covers the rebind-discovery need.

## Verification
- Full suite **1094 examples, 0 failures**; typecheck + `crystal tool format` clean (except the two files the formatter pre-existingly crashes on); no new ameba findings.
- New specs: palette reflects a rebind + ignores FIXED-id hand-edits; registry collision detection; SpaceMenu grows past 12 + shows the scroll marker when clipped; Miner rename label.
- Live E2E (tmux, real proxy traffic): `j/k` scroll the Replay response, palette renders chords, Help shows the `n` entry, and the History space menu now shows all 13 entries with a `▼` cue when a short terminal clips it.

### DIFF
--- a/spec/tui/miner_view_spec.cr
+++ b/spec/tui/miner_view_spec.cr
@@ -12,4 +12,14 @@ describe Gori::Tui::MinerView do
       config: json, flow_id: nil, position: 0, name: nil))
     restored.config.notify.should eq(Gori::Miner::NotifyMode::WhenFound)
   end
+
+  it "fronts a custom name on the sub-tab label (the rename path now wired for Miner)" do
+    view = Gori::Tui::MinerView.new
+    view.load("http://h.test", "GET /a HTTP/1.1\r\nHost: h.test\r\n\r\n".to_slice, false, nil,
+      Gori::Miner::Config.new)
+    auto = view.label(18)
+    view.name = "login probe" # MinerController#apply_rename sets this
+    view.label(18).should contain("login")
+    view.label(18).should_not eq(auto) # the custom name overrides the derived summary
+  end
 end
--- a/spec/tui/palette_spec.cr
+++ b/spec/tui/palette_spec.cr
@@ -118,6 +118,50 @@ describe Gori::Tui::PaletteState do
     ids.should contain("import.oas")
   end
 
+  it "renders a verb's EFFECTIVE chord so a rebind is reflected (not the default)" do
+    prev = Gori::Settings.keymap_overrides
+    begin
+      ctx = FakeExecContext.new
+      palette = PaletteState.new(Gori::Verbs.registry)
+      palette.reset(ctx)
+      "Toggle capture".each_char { |c| palette.append(c, ctx) } # capture.toggle (default: c)
+
+      # Default binding: the rebound label is nowhere on screen yet.
+      Gori::Settings.keymap_overrides = {} of String => Array(String)
+      base = MemoryBackend.new(80, 24)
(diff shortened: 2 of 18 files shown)


hahwul:
## Review (8 finder angles × verify, medium effort)

**No correctness bugs.** All surviving findings are maintainability/quality; the two candidates that questioned the `^B ws` hint removal were REFUTED (confirming the removal is correct — `^B` only flips `@reveal`, which has no visible effect in those editors).

**Applied in this PR:** extracted `Hotkeys.rebindable_overrides(registry)` so the palette and `build_keymap` share one override-filter instead of two byte-identical inline copies — the duplication was the sole guarantor of "palette shows what dispatch uses," so sharing it removes the drift risk this PR set out to prevent.

**Deferred as follow-ups** (out of this PR's scope or broad refactors, none blocking):
- Scroll-marker `↕/▲/▼` glyph mapping is now duplicated in 4 views (`settings_view`, `hotkeys_overlay`, `setup_wizard`, `space_menu`) — a shared helper would fold them, but 3 sites predate this PR.
- The `key.up?, key.lower_k?` / `key.down?, key.lower_j?` alias pairing recurs in ~10 handlers — a `nav_delta`/`Keys.up?` helper could centralize it.
- `space_menu.cr` `visible` param is derivable (`last == i == rows-1`); minor.
- `subtab_new` and `subtab_new_supported?` are two hand-maintained `{replay,fuzzer,convert,notes}` lists (linked by a comment) — low drift risk today.

PR #110, opened 2026-07-10
fix(prism): 25+ correctness/robustness bugs across active, passive, TUI, orchestration

hahwul:
## Summary

A multi-round audit of the **Prism** scanner (the least-tested feature) — active probes, passive rules, orchestration, the TUI, and the store/CLI/MCP surface. Each finding was surfaced by a fan-out of focused reviewers, **adversarially verified** (default-refute) to drop plausible-but-wrong reports, then reproduced against real code before fixing. ~25 confirmed bugs across 5 commits, every fix with regression coverage.

## What's fixed, by area

**Active scan** (`d944da1`)
- Queued canary/CORS probes kept firing at the target after the operator switched Active→Passive/Off (`run_active` only gated on `@stopped`; `set_mode` can't drain the queue). **Medium.**
- `origin_form` dropped the query of a pathless absolute-form target (`GET http://h?q=v` → `/`) so its params were never probed, and mistook a `/` inside the query for the path. Extracted one correct shared helper.
- `CorsReflection` sent an absolute-form request line direct to the origin (some origins reject it → missed reflections). Normalized to origin-form.
- A merged `(code,host)` group escalated severity but froze the title, so a `reflected_param` row showed a MED badge next to the non-HTML title. Title now tracks the highest-severity observation (in both the store and the headless `group.cr`).

**Passive rules** (`70e502d`)
- CSP duplicate-directive resolution was last-wins vs. the browser's first-wins (weak_csp FP **and** FN).
- Logout cookies with a sentinel value (`sid=deleted; Max-Age=0`) or a negative Max-Age tripped hygiene findings.
- A Basic challenge listed after another scheme in one `WWW-Authenticate` header was missed.
- PGP (`… PRIVATE KEY BLOCK`) and PKCS#8 `ENCRYPTED PRIVATE KEY` blocks never matched the secret pattern.
- A 4-part software version (`File version 10.0.1.2`) was flagged as a private-IP leak.
- GraphQL introspection used loose substrings (echoed query FP; result past the 64 KiB cap FN) → anchored on the `"__schema":{` result envelope.

**Orchestration** (`9ba90f9`, `158c933`)
- Replay/CLI/MCP-sourced scans lost **all** request headers (editor stores bare-LF, parser needs CRLF), silently missing CORS/Basic-auth/tech rules. **Medium.**
- WebSocket rescans re-scanned the whole buffer per frame (hit_count inflation, O(n²)); reworked to a per-flow high-water-mark that pages forward and scans each frame exactly once — fixing both an evicted-flow re-count and a >WS_MSG_CAP backlog skipping a band of frames (missed secret). **Medium.**
- A burst overflowing the droppable `prism_events` feed permanently skipped a plain-HTTP flow's only scan trigger; added a periodic passive catch-up sweep.
- A probe dropped by the mode gate kept its dedup key, suppressing re-probing after Active is re-enabled — key now released.

**TUI** (`d54ee08`)
- Bulk "dismiss all with this code" ignored the ⇧S scope lens (muted out-of-scope hosts, over-reported the count). **Medium.**
- Delete confirmed one issue but deleted whatever was selected at confirm time (a background reload could diverge them). Now targets the captured id. **Medium.**
- Promote created duplicate Findings and dropped Replay-only evidence.
- Narrow-terminal rendering: preview focus stranded list nav, preview scroll never upper-clamped, severity tallies overpainted the MODE chip.

## Not changed
- **store / filter / CLI / MCP** round: adversarial verification confirmed **0** real bugs — the layer is solid.
- **Known limitation documented in-place:** a still-streaming WebSocket flow can be reaped by retention after 100k newer flows; a correct fix needs a WS-relative recency floor (the flow vs. ws-message timestamps are set from different sources) and belongs in a focused retention change.

## Verification
- 180 Prism-touching examples pass; full build (`crystal build src/main.cr`) succeeds.
- The 4 pre-existing `store_spec.cr` failures (migration/FTS/h2 tests) are unrelated — identical on the pristine tree (verified via `git stash`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/spec/prism_spec.cr
+++ b/spec/prism_spec.cr
@@ -285,6 +285,33 @@ describe Gori::Prism::Analyzer do
     end
   end
 
+  it "pages a >WS_MSG_CAP WebSocket backlog without skipping a band (no missed secret)" do
+    with_store do |store|
+      detail = capture_flow(store,
+        "HTTP/1.1 101 Switching Protocols\r\nUpgrade: websocket\r\nConnection: Upgrade\r\n\r\n",
+        target: "/ws", status: 101, content_type: nil,
+        req_headers: "Upgrade: websocket\r\nConnection: Upgrade\r\n")
+      fid = detail.row.id
+      store.insert_ws_message(fid, "in", 1, "hello".to_slice) # frame 1 (no secret) → sets hwm
+      scope = Gori::Scope.load(store)
+      feed = Channel(Gori::Store::FlowEvent).new(8)
+      a = Gori::Prism::Analyzer.new(store, scope, feed, Gori::Prism::Mode::Passive, true)
+      a.start
+      feed.send(Gori::Store::FlowEvent.new(fid, :updated)) # initial scan; hwm = frame 1
+      sleep 120.milliseconds
+      # A burst of >WS_MSG_CAP(200) frames arrives with the secret in frame ~30 — the band a
+      # last-200-window rescan would drop (window would be frames ~52..251).
+      250.times do |k|
+        payload = k == 28 ? "token=AKIAIOSFODNN7EXAMPLE" : "frame#{k}"
+        store.insert_ws_message(fid, "in", 1, payload.to_slice)
+      end
+      feed.send(Gori::Store::FlowEvent.new(fid, :updated)) # one rescan must page the whole backlog
+      sleep 250.milliseconds
+      store.prism_issues.count(&.code.== "secret_in_ws").should eq(1) # the banded secret was caught
+      a.stop
+    end
+  end
+
   it "bumps store.prism_generation on persist and honors session suppress after hard delete" do
     with_store do |store|
       detail = capture_flow(store, "HTTP/1.1 200 OK\r\nContent-Type: text/html\r\nServer: x\r\n\r\n",
--- a/spec/prism_spec.cr
+++ b/spec/prism_spec.cr
@@ -312,6 +312,30 @@ describe Gori::Prism::Analyzer do
     end
   end
 
+  it "scans a WebSocket flow FIRST seen with a large backlog from its oldest frame" do
+    with_store do |store|
+      detail = capture_flow(store,
+        "HTTP/1.1 101 Switching Protocols\r\nUpgrade: websocket\r\nConnection: Upgrade\r\n\r\n",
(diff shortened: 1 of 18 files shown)


hahwul:
## Code review (2 independent adversarial reviewers) — addressed in `7981566`

**Store / passive changes:** verified **correct**, no defects. `prism_issues.title` is `TEXT NOT NULL` (never added by a later migration), so the non-null title read can't crash on a legacy row; the title-sync tie/downgrade logic and dynamic Info-title tech codes are unaffected; `ws_messages_after` SQL/bind/column order is correct; the `from_replay` LF→CRLF `gsub` is idempotent and head-only; `group.cr` agrees with the store; `security_headers` `||=` first-wins is correct; `cookies` `max_age_delete?` is a strict superset of the old regex; `graphql` `\s*` covers the newline variants.

**Analyzer concurrency:** verified safe — no double full-scan between `catch_up` and `passive_loop` (the `@analyzed` check→claim window contains only non-yielding SQLite reads), paging terminates correctly, and teardown/use-after-close is guarded (`@prism.stop` before `@store.close`, rescues in place).

**Three real issues found in this PR's own Round-5 WS changes — fixed:**
1. *(Medium)* `rescan_ws`'s first (no-hwm) scan grabbed only the **last** window, so a 101 flow first scanned late (via `catch_up`) with a >`WS_MSG_CAP` backlog permanently skipped its oldest band → **now always pages forward from the hwm (0 on first scan)**, covering every frame from frame 1. Added a regression test.
2. *(Low)* `@ws_hwm` eviction was FIFO not LRU, so a long-lived active socket could be evicted first (>10k WS flows) then re-scanned → **delete+re-insert on update** keeps active flows at the back of the eviction order.
3. *(robustness)* the private-IP scan's `m.pre_match` allocated the whole prefix per match (O(n²) transient on a crafted version-heavy body) → **bounded 24-char window off the match index**.

Two non-defects left as-is: the `auth` comma-split has a contrived Digest-realm-contains-`, basic ` FP already noted in the code comment (the fix is a clear net win), and one documented retention-prune limitation for still-streaming WebSockets.

Verification: 101 Prism-touching + 586 TUI/replay/scope/fuzz examples pass; full build succeeds.

Issue #356, opened 2026-07-25
refactor(tools): extract per-tool Plan builders so TUI/CLI/MCP share run assembly


hahwul:
## Problem

Every multi-surface tool re-implements its "assemble a run" pipeline once per surface. For the fuzzer, the sequence *template parse → auto_mark → payload sets → matcher → config → generator → sender → scope → engine* exists three times:

- `src/gori/tui/fuzzer_view.cr:928` — `build_engine`
- `src/gori/cli/run/fuzz.cr:100-121`
- `src/gori/mcp/tools/fuzz.cr:221-250` — `build_fuzz_job`

The same shape repeats for miner, sequencer, discover, and repeater/minimize.

This is not merely duplication — **the copies have diverged on semantics**:

```crystal
// CLI (run/fuzz.cr:116) — snapshot Scope, up-front guard, no reload; nil scope ⇒ NO gate
scope = cli_scope(project_name, db_path, flow_id)
guard_active_scope(scope, scheme, host, request_target(text.to_slice), allow_unscoped, "gori run fuzz")
backend = scope ? Fuzz::ScopedBackend.new(sender, scope) : sender

// MCP (tools/fuzz.cr:246) — private Scope, periodic reload
backend = Fuzz::ScopedBackend.new(sender, Scope.load(store), reload_every: Fuzz::ScopedBackend::RELOAD_INTERVAL)

// TUI — shared @session.scope, refreshed in place by the data_version poll
```

Each is individually reasoned in comments, but the user-visible result is that the same tool behaves differently depending on how it was invoked. `HostOverrides` loading, `Env.expand` application, and `Decoder.shared_registry` wiring diverge across the three in the same way.

The missing layer is a normalized-options → ready-engine builder. Today each surface owns both option parsing *and* assembly; only option parsing is legitimately surface-specific.

## Proposed direction

For each tool, introduce a `Plan` builder in the tool's own module (core layer):

```
Fuzz::Plan.build(options : Fuzz::PlanOptions, outbound : Outbound) : Fuzz::Engine
```

- `PlanOptions` is a plain struct: template source, marks, payload sets, processors, matcher config, mode, concurrency/rps/retries/timeout, origin, http2, evidence policy
- Each surface's job shrinks to *parse its own input format into `PlanOptions`* — CLI from `OptionParser`, MCP from the JSON hash, TUI from view state
- Assembly, `Env.expand`, host overrides, decoder registry, and the scope decision live in exactly one place
[...]

PR #361, opened 2026-07-25
refactor(scope): single outbound chokepoint for the active-traffic scope gate (#354)

### LINKED ISSUE #354: refactor(scope): single outbound chokepoint for the active-traffic scope gate (~20 scattered sites)
## Problem

The security-critical invariant "no active request leaves gori without a scope decision" is enforced by **convention at ~20 separate call sites across 3 surfaces**, with no chokepoint.

Sites matching `guard_active_scope|scope_check|ScopedBackend|SCOPE_BLOCKED`:

| Surface | Files (count) |
|---|---|
| MCP | `tools/send.cr` (5), `tools/minimize.cr` (5), `tools/mine.cr` (3), `tools/fuzz.cr` (3), `tools/probe.cr` (2), `tools/sequence.cr`, `tools/discover.cr`, `server.cr` |
| CLI | `run.cr` (3), `run/fuzz.cr` (3), `run/sequence.cr` (2), `run/repeater.cr` (2), `run/mine.cr` (2), `run/repeater_minimize.cr` (2) |
| TUI | `fuzzer_view.cr` (2), `miner_view.cr` (2), `controllers/repeater_controller.cr` (2) |
| core | `probe/active.cr` (2), `probe/scan.cr`, `fuzz/engine.cr` |

Adding a new active feature requires the author to *remember* to gate it on every surface. That has been missed repeatedly:

- Repeater dialled direct with no scope gate on TUI `^R` / group send / WS / Minimize **and** `gori run repeater` (MCP `send.cr` was already correct — so the gap was per-surface, not per-feature)
- fuzz / mine / sequence CLI gating mirrored `discover`'s CLI path but not its MCP path

## The deeper issue: the three surfaces disagree on semantics

Same tool, three different scope lifetimes:

```crystal
// src/gori/cli/run/fuzz.cr:116 — snapshot Scope + up-front guard, no reload

// src/gori/mcp/tools/fuzz.cr:246 — private Scope, periodic reload

```

Each has a documented rationale, but the result is that a mid-run EXCLUDE or Sandbox toggle is honoured on MCP, ignored on CLI, and honoured-by-side-effect on TUI. `scope ? ScopedBackend : sender` on the CLI path also means **no project ⇒ no gate at all**, which is the fail-open direction.

## Proposed direction

Make the scope decision a **required constructor argument**, so an ungated sender cannot be built:

1. Introduce one seam (e.g. `Gori::Outbound`) that every active sender (`Fuzz::Sender`, `Miner`, `Sequencer`, `Repeater::Engine`/`H2Engine`/`WsEngine`, `Probe::Active`) must be constructed with.
2. Model the decision explicitly rather than as `Scope?`:
   `Gated(scope)` | `Unscoped(reason)` — where `Unscoped` is only constructible from an explicit `--allow-unscoped` / `allow_unscoped:true`. A missing project becomes `Unscoped(:no_project)`, a *deliberate* state rather than a nil that silently skips the gate.
3. Pick ONE reload semantic and apply it everywhere (recommend MCP's `reload_every` — it is the only one that honours a mid-run policy change; document the choice in `DESIGN.md` under P5).
4. Delete the per-surface `guard_active_scope` / `scope_check` duplicates; the seam does the up-front check once.

## Acceptance criteria

- [ ] No active sender can be constructed without an `Outbound` decision (compiler-enforced, not convention)
- [ ] `grep -rn 'guard_active_scope\|scope_check' src/` returns only the seam's own definition
- [ ] Sandbox / EXCLUDE semantics are byte-identical across TUI, `gori run`, and MCP for: repeater send, fuzz, mine, sequence, discover, probe active, minimize
- [ ] Spec: a table-driven test asserting each (surface × tool) pair refuses an out-of-scope target, and that `allow_unscoped` is the only way through
- [ ] Spec: a mid-run EXCLUDE is honoured on all three surfaces

## Parallelism

- **Blocks #356** (per-tool Plan builders). #356 extracts the shared run-assembly pipeline, which must consume this seam — if #356 lands first it will bake the three divergent scope semantics into the new builder and #354 rewrites it immediately. Land #354, then #356.
- **Parallel-safe with #353, #355, #358.** Overlap with #355 is limited to `tui/fuzzer_view.cr` / `tui/miner_view.cr` (this issue) vs `tui/runner.cr` + `*_overlay.cr` (#355) — disjoint files.

hahwul:
Closes #354.

## What

The invariant **"no active request leaves gori without a scope decision"** was enforced by convention at ~20 call sites across three surfaces, with no chokepoint. Adding a new active feature meant remembering to gate it on every surface, and that had been missed repeatedly.

`Gori::Outbound` (`src/gori/outbound.cr`) makes the decision a **required constructor argument** on the two active senders, so an ungated sender is a compile error rather than a security hole:

| Sender | Covers |
|---|---|
| `Fuzz::Sender` | fuzz, mine, sequence, repeater minimize, probe active — TUI + CLI + MCP |
| `Repeater::Sender` (new) | TUI `^R` / send-group / WS, `gori run repeater`, MCP `send_request` / `send_websocket` |

`Fuzz::ScopedBackend` and the per-surface `guard_active_scope` / `scope_check` / `sandbox_block_reason` duplicates are deleted. `grep -rn 'guard_active_scope\|scope_check' src/` now returns nothing.

## The two layers

- **Layer 1 `#check`** — the include/allowlist gate, made once up front. Its strictness is the *one* thing that legitimately varies per surface, so the variants are **named** rather than re-derived at each site: `Outbound.agent` (MCP: refuse anything not included, incl. an unconfigured project), `Outbound.cli` (`gori run`: an unconfigured project stays permissive), `Outbound.interactive` (TUI: no up-front gate, the operator typed the target).
- **Layer 2 `#sweep_block` / `#send_block`** — Sandbox always, plus explicit EXCLUDE for an automated sweep. Identical on every surface, and applied **even when Layer 1 was waived**. The sweep-vs-single-send split is the existing, documented one (`Interceptor#sandbox_blocks?`): an EXCLUDE stops a sweep but not one hand-authored Repeater send. No existing gate was weakened to make the refactor tidy.

Waiving Layer 1 is reachable only from an explicit `--allow-unscoped` / `allow_unscoped:true` or a genuinely absent project, and yields a named `Unscoped(reason)` visible in the audit line. The CLI's `scope ? ScopedBackend.new(sender, scope) : sender` fail-open becomes `Unscoped(:no_project)` — a deliberate state, not a silent skip.

## Reload semantics: MCP's `reload_every`, everywhere

Recorded in `DESIGN.md` §7 (refines **P5**), and §3's "see issue #354" open question is closed out. Before: MCP re-read the rules on a throttled interval, `gori run` snapshotted at start-up and closed the store so it *could not* re-read, and the TUI re-read only as a side effect of its `data_version` poll. A mid-run EXCLUDE stopped an MCP sweep, was invisible to a CLI sweep, and stopped a TUI sweep by accident.

`Outbound` now owns the reload for every surface, throttled to `RELOAD_INTERVAL` (1s), clock advanced *before* the blocking reload so worker fibers can't stampede the store, failures swallowed so the last-known rules stay in force. On the CLI this means the read connection lives as long as the run (`Outbound#close` releases it) instead of being closed right after the snapshot.

## Gaps this closes (each previously invisible)

- **TUI Sequencer** built a bare `Fuzz::Sender` with **no gate at all** — Sandbox did not contain it.
- **`Probe::Analyzer`** (the TUI's live active prober) had no Layer 2, unlike `Probe::Scan` on CLI/MCP.
- **MCP `send_request` / `send_websocket`** let `allow_unscoped:true` walk straight past Sandbox.
- **MCP `sequence_start`** had only the job-start check, no per-send gate.
- Both layers now anchor the URL on the host actually **dialled** (`Outbound.scope_url`). Only MCP `send_request` did that before, so a spoofed absolute-form request line (`GET http://acme.test/x` while dialling `evil.test`) could be judged against the wrong host's rules — an anchored include for `acme.test` would authorise the send.

## Specs

`spec/outbound_spec.cr` (20 examples) pins the (surface x layer) table, that `allow_unscoped` is the only way through Layer 1, that Sandbox holds under it on every surface, that a mid-run EXCLUDE is honoured on all three, and the dial-host anchoring.

## Scope note for the reviewer

Two files outside the issue's stated boundary were touched, both unclaimed by sibling agents and both required by compiler enforcement: `src/gori/tui/sequencer_view.cr` (+ a one-line signature change in `controllers/sequencer_controller.cr`) and `src/gori/probe/analyzer.cr` — those are exactly the two ungated senders listed above. `src/gori/discover/**` is deliberately untouched; Discover keeps its own equivalent gate in `discover/adapters.cr`.

## Follow-up

#356 (per-tool Plan builders) is unblocked by this. The API shape it should consume is `Plan.build(options, outbound)` — see the comment I'll leave on that issue.

## Verification

- `just test` — 3706 examples, 0 failures
- `just check` — format clean on touched files; ameba at or below the pre-existing baseline for every touched file (two new files are clean)


hahwul:
## Code-review pass (required gate) — results

Ran the review with two adversarial specialists (security + behavior-parity) plus my own pass. **Four real findings, all fixed before merge.** Both specialists independently converged on the same two bugs, which is a good signal they were real.

### Fixed

1. **`gori run repeater send|minimize` silently lost the Sandbox gate when `--project`/`--db` were omitted** (self-found, both specialists confirmed). `cli_outbound` short-circuited on `unless flow_id || project_name || db_path`, and the repeater callers pass `flow_id: nil` — so the common no-flag invocation got `Unscoped(NoProject)`, which waives *both* layers. `origin/main` unconditionally loaded the scope off `resolve_read_project`. Split into `optional_project_outbound` (fuzz/mine/sequence only, which can genuinely run project-less via `--request`/stdin) and `project_outbound`; the name now makes a wrong pick visible at the call site. Regression spec in `spec/outbound_spec.cr`.

2. **Layer 2 gated on the spoofed request-line host for absolute-form requests** (self-found, both specialists confirmed). `Scope.request_url` returns an absolute-form target verbatim, so `GET http://acme.test/x HTTP/1.1` dialled at `evil.test` was judged against `acme.test`'s rules — an anchored include for `acme.test` would authorise the send. This is exactly the bypass MCP's `request_scope_url` was written to prevent, and it only ever protected MCP `send_request`. Hoisted into `Outbound.scope_url`, so all three surfaces now anchor on the dial host. Spec'd.

3. **`calibrate_baseline` sat outside the block that releases the CLI read connection** — it sends, so a raise there leaked the handle. Moved inside.

4. **DESIGN.md overstated the reload uniformity.** A one-shot Repeater send builds a fresh decision per send, so the throttle window never elapses within it and no reload fires. Reworded: the reload is uniform for long-running jobs, which is where the three surfaces actually diverged.

### Raised and deliberately not fixed here

- **`gori run fuzz|mine|sequence --request` with no project flags waives both layers.** Pre-existing (`origin/main` did the same via `scope ? ScopedBackend : sender`) and consistent with `cli_host_overrides`, which skips on the identical condition. But `resolve_read_project(nil, nil)` *does* resolve a default project, so the active project's Sandbox genuinely does not apply to that run. Binding a standalone run to the ambient project is a **product decision**, so I did not make it unilaterally — instead the run now prints a one-line STDERR notice so an ungated run stops looking gated. Worth a follow-up issue.

- **Discover's seed + `robots.txt`/`sitemap.xml` bypass its own Layer-2 gate** (`discover/engine.cr#seed_frontier` pushes straight past `bounded_url`). Blast radius 3 requests; everything after the seed is gated. Pre-existing, and `src/gori/discover/**` is outside this issue's file boundary — `git diff origin/main -- src/gori/discover/` is empty. Follow-up.

- **`proxy/conn/client_conn.cr:1146` hard-codes `"blocked by sandbox (out of scope)"`** instead of `Outbound::SANDBOX_ERROR`. Same string today; it's a proxy hot-path file outside this boundary, so I left it. Trivial follow-up.

- **Minimize backend stacking flipped** `Scoped(Capped(Sender))` → `Capped(Sender-gated)`, so a blocked send now consumes a `SEND_CAP` slot. Traced through `Repeater::Minimize.run`: a blocked target fails calibration identically either way, and the divergence needs a >250-candidate session blocked mid-run, which was unreachable before (minimize had no mid-run reload at all). Not observable.

### Verified clean by the parity pass

Error strings byte-identical (and nothing string-compares them); `ScopeCheck` → `Outbound::Verdict` preserves the JSON contract on all five emitters; `Gate::Configured` / `Gate::Allowlist` reproduce `guard_active_scope` / `scope_check` / `active_target?` exactly; backend stacking for fuzz/mine/sequence keeps the same relative order; `send_group`/`send_ws`/`send` pass-through defaults reproduce the old omitted-argument calls (including the pre-existing TUI-doesn't-pass-`overrides` asymmetry, preserved not changed); no store deadlock from the overlapping WAL read connections, and every write-store open happens after `outbound.close`.

The security pass also swept the whole tree for `Upstream.dial*` / `TCPSocket.new` / `Engine.send` / `send_pipeline` / `WsEngine.send` and confirmed every active dial is now behind `Fuzz::Sender`, `Repeater::Sender`, or `Discover::Sender`.

### Verification

- `just test` — 3706 examples, 0 failures
- CI green on the first commit (build + spec on Crystal 1.20.2 and 1.21.0, both docker builds); re-running for `e118828`
- ameba at or below the pre-existing baseline for every touched file; the two new files are clean

hahwul:
Coordinator: the three residual gaps from your code-review pass are now tracked so they survive the merge.

- #363 — `gori run --request` with no project flags waives both layers (product decision; three options laid out)
- #364 — Discover's seed + `robots.txt`/`sitemap.xml` bypass Layer 2 (3 requests)
- #365 — `client_conn.cr:1146` hard-codes the sandbox string instead of `Outbound::SANDBOX_ERROR` (good first issue)

All three independently verified against `origin/main`. Declining to fix them out of lane was the right call — merge when CI is green.

PR #366, opened 2026-07-25
refactor(fuzz): one Plan builder for TUI/CLI/MCP run assembly

### LINKED ISSUE #356: refactor(tools): extract per-tool Plan builders so TUI/CLI/MCP share run assembly

hahwul:
Closes #356 (fuzz only — miner / sequencer / discover / repeater follow as separate PRs, per the issue's "one agent per tool" plan).

## What

`Fuzz::PlanOptions` (a plain struct) + `Fuzz::Plan.build(options, outbound) : Fuzz::Plan` replace the three hand-written copies of the fuzzer's assembly pipeline. Each surface now contains only option parsing:

| surface | before | after |
|---|---|---|
| `tui/fuzzer_view.cr#build_engine` | parse → mark → sets → generator → sender → engine | view state → `PlanOptions` |
| `cli/run/fuzz.cr` | same, 20 lines | `OptionParser` → `PlanOptions` |
| `mcp/tools/fuzz.cr#build_fuzz_job` | same, 25 lines | args hash → `PlanOptions` |

`grep -rn 'Fuzz::Engine.new' src/` → no matches; the only construction is `Engine.new` inside `src/gori/fuzz/plan.cr:182`.

## The `Outbound` stays an argument

Layer-1 strictness differs per surface **on purpose** (`Outbound.agent` / `.cli` / `.interactive`, DESIGN.md §7). `Plan.build` never constructs one — doing so would collapse a deliberate distinction into whichever policy got hard-coded.

## Divergences settled (the issue's "pick the correct behaviour and say which")

1. **`Env.expand` runs once**, on the raw template and on the resolved target. Both `gori run fuzz` and MCP expanded a seeding flow's target *twice* (once directly, again inside `resolve_fuzz_target` / `fuzz_origin`). Expansion is a single pass, so a second pass resolves tokens the first pass *produced* — a var whose value contained a `$TOKEN` behaved differently per surface.
2. **The Layer-1 gate string comes from the template's baseline rendering**, not its raw first line. The TUI's template arrives already marked, so the raw line fed `/find?term=§VAL§` into the gate there while feeding `/find?term=VAL` from the CLI and MCP. A `§` in a *path* position defeats an anchored include rule. Rendering each position's own default back out is marker-free on all three and byte-identical to what the CLI/MCP previously passed.
3. **`Decoder.shared_registry` is wired inside the builder** rather than passed by each caller. All three already did it; now a new surface cannot forget it and silently send un-transformed payloads.

**HostOverrides is the one gap left open.** The TUI has never applied them to a fuzz send (the proxy path does, via `Proxy::Upstream.dial`). `PlanOptions#overrides` carries them and `build_engine` now takes the parameter, but passing `session.host_overrides` is a `fuzzer_controller.cr` edit — outside this PR's file boundary. Behaviour is unchanged; the fix is one call site.

## Errors stay in each surface's vocabulary

`Fuzz::PlanError` carries a machine-readable `reason`; every surface renders its own sentence (`--auto` / `auto:true` / `^A params`). Sharing the assembly must not flatten three vocabularies into one.

## Verification

Beyond `just check` + `just test` (3903 examples, 0 failures) and the new `spec/fuzz/plan_spec.cr` (14 examples), all three surfaces were driven for real:

- **CLI** — every abort path (`no positions` / `no payloads` / `--target is required` / `could not determine a target host` / `unsupported target scheme`) byte-identical to `main`; live sniper + cluster-bomb sweeps with `--mark`, `--prefix` and the mark-occurrence warning against a local server; scope gate confirmed allowing in-scope, refusing out-of-scope, and honouring `--allow-unscoped`.
- **MCP** — driven over real stdio JSON-RPC (`gori mcp`): `fuzz_start` → `fuzz_results` with processors, plus every error message and the `SCOPE_BLOCKED` refusal, all matching `main`.
- **TUI** — Fuzzer tab under tmux: both error hints and a real 2-request sweep with payloads substituted at the marked position.

The gate fix in (2) was confirmed by an accidental A/B: a stale binary carrying `main`'s raw-line behaviour **blocked** `/exact/§x§` against `^http://127\.0\.0\.1/exact/x$`; the rebuilt one allows it.

## Notes for review

Two error-precedence shifts, both cosmetic and both between messages that are individually unchanged:
- CLI: the "no project in play" STDERR notice now precedes a usage-error abort (the `Outbound` is built before `Plan.build`).
- MCP: argument-shape errors (malformed `marks` / `payloads` / regex) now surface before "template has no §…§ positions".

`.presence` now applies to a flow's `default_target` as well as MCP's `url`, so a blank flow target yields `NoTarget` rather than `BadTarget`.


hahwul:
Coordinator review — acceptance criteria met, and the design went further than the brief asked.

Verified:
- `grep -rn 'Fuzz::Engine.new' src/` → only `src/gori/fuzz/plan.cr:182`. The Discover/Miner/Sequencer hits are their own engines, correctly out of scope for this PR.
- The assembly genuinely moved. `fuzzer_view.cr#build_engine` went from 12 lines of template/sets/generator/sender/matcher wiring to `PlanOptions` construction plus one `Plan.build` call.
- `Plan.build` takes the `Outbound` as an argument rather than constructing one, so the deliberate per-surface Layer 1 distinction (`agent`/`cli`/`interactive`, DESIGN.md §7) survives.

The `PlanError::Reason` enum was not asked for and is the right call: the builder reports a machine-readable reason and each surface words it in its own vocabulary, which is why the TUI can name `^A`/`^K`/`^O`/`^L` in its hint while CLI and MCP say something appropriate to themselves. The exhaustive `in` match means a new reason cannot be silently unhandled on a surface.

On the hostname-overrides note in your `build_engine` comment: the gap is broader than the Fuzzer tab. **All five** TUI active senders omit `overrides` while all six CLI/MCP equivalents pass it — `fuzzer_view.cr:943`, `miner_view.cr`, `sequencer_view.cr`, `controllers/repeater_controller.cr:1409`, `controllers/discover_controller.cr`. Tracked as #367 with the full table. Keeping the parameter you added is right; leave the other four alone here.

Merge when CI is green.

PR #370, opened 2026-07-25
refactor(tui): migrate the rule, provider and import modals onto the Overlay seam (#355)

### LINKED ISSUE #355: refactor(tui): finish the Overlay migration — 28 of 31 modals still ride the Runner ladder
## Problem

```crystal
```

| Ladder | Line | Generic? |
|---|---|---|
| IME preedit | 933 | yes |
| key | 1048 | yes |
| click | 1403 | yes |
| wheel / move | 1672 | yes |
| render | 3798 | yes |

|---|---|---|

### Out of scope

## Target on completion

- `Runner` ivars under ~40

## Parallelism

hahwul:
Phase 1 **batch C2** of #355. Migrates `ca_import`, `import`, `oast_provider`, `custom_rule` (probe_rule) and `rewriter_rule` onto the polymorphic `Overlay` seam.

`runner.cr` loses 227 lines: five modals' arms in all seven dispatch ladders (preedit / key / capture-gate / click / wheel / title / hint), 19 private handler methods, and six ivars (`@import_overlay`, `@ca_import_overlay`, `@custom_rule_overlay`, `@rewriter_rule_overlay`, `@oast_provider_overlay`, `@rewriter_preview_sig`). Domain coupling moves to an `on_commit` closure at each open-site, mirroring `ScopeRuleOverlay`.

## Two pre-existing bugs that fall out of the move

**1. `probe_rule` had no chrome at all.** It was missing from *both* the `focus_label` and `key_hints` ladders on `main` (the other four modals were in both). With the custom-rule form open, the badge showed the tab body's label and the bottom row advertised `^N new` / `^W close` while a modal owned the keyboard. `title`/`hint` are abstract on `Overlay`, so migrating closes the gap by construction. **This is the one place the diff is not byte-for-byte behaviour-preserving** — deliberately. The other four modals' title/hint strings are byte-identical to their deleted ladder arms.

**2. `Overlay#title` silently collides with `CustomRuleOverlay#title`** — the rule's own user-entered title, which `ProbeController` reads to persist the rule. Crystal has no `override`, so this is silent in *both* directions: left alone the badge shows the rule name; "fixed" naively, every saved rule is named `"CUSTOM RULE"`. The form's accessor is now `rule_title` (following the precedent `OastProviderOverlay#provider_name` already set); 4 call sites in `probe_controller.cr` updated. Caught by a spec, then proved end-to-end in the running TUI.

## The CA-import ordering trap

`submit_ca_import` is the **one** commit closure that closes itself and returns `false`. The shell's post-commit `close_active_overlay` unconditionally resets `@overlay`, which would wipe the destructive-CA confirm this opens one line after opening it — the trap recorded on #355 as directly blocking C2. Every other closure returns `true` and lets the shell close it; the asymmetry is commented at both sites. Verified in the running TUI: the confirm actually appears.

## Rewriter preview

The "only rescan when a match-relevant field changed" gate is preserved, but split by dependency: the **gate** (form state) moves into the overlay, while the **scan** (a traffic read) is injected as `on_preview` at the open-site. `Runner#refresh_rewriter_preview` and `@rewriter_preview_sig` are gone. Without the gate, every keystroke would re-run a 200-flow scan.

## Shared spec file

`MODAL_OVERLAYS`'s assertion becomes derived set-complementarity instead of `unmigrated.size.should eq(28)`. That literal was one line five parallel batches each wanted a different integer on, and a conflict resolved by keeping your own value asserts a total that is wrong for the combined state. It also could never catch an accidental **extra** member. The per-batch edit is now appending to one `MIGRATED_KINDS` list. Both failure directions proved by mutation.

## Verification

Specs alone can't reach the Runner (it needs a tty), so every modal was also driven in a real TUI under tmux:

| modal | verified in the running app |
|---|---|
| ca_import | badge/hint; empty-path keeps the form up + toasts; filled → the destructive confirm appears |
| import | badge reads `IMPORT HAR` (per-kind, not generic); a real HAR imported; card closed; toast |
| custom_rule | badge `CUSTOM RULE` (new); rule persisted as `leaky header`, **not** `CUSTOM RULE` |
| rewriter_rule | badge/hint; live preview both states; rule saved and live in the preview output |
| oast_provider | badge/hint; saved with **global** scope preserved |

- `crystal spec spec/tui` → **1162 examples, 0 failures**
- `crystal spec` → 3948 examples, 8 failures, **all pre-existing/environmental** (a `gori` holding :8070); reproduced identically at `origin/main` with no patch applied
- Re-verified in a clean isolated worktree (`origin/main` + this diff only)
- **Six mutations, six failures** — no spec passes vacuously: `rule_title` collision reintroduced, preview gate removed, ca_import committing from any row, import title losing its per-kind label, and an extra/missing `MODAL_OVERLAYS` member
- ameba: baseline unchanged (one finding *removed*, none added); format clean on every touched file

## Deliberately not done here

- Extracting a `RowFormOverlay` base for the 6 now-identical row-form `handle_click` bodies — touches `overlay.cr` + `scope_rule_overlay.cr`, both outside this batch while four agents are in flight. Phase 2.
- Renaming the contract method `title` → `badge_title`, which would remove the collision class for good (`IssueNew` is the next landmine). Phase 2.
- One stale orphan comment at `runner.cr:1759` — sits adjacent to C1's deletion range, not worth the conflict.

Composes with C5's `on_close` base change: this branch does not touch `overlay.cr`, and my overlays set no `on_close`, so `close_active_overlay` behaves identically under it.

Part of #355 — batch C2 of 5. Deliberately NOT "Closes #355": C1, C3, C4 and C5 are still in flight, and an auto-close here would drop the tracking issue four batches early.

### DIFF
--- a/spec/tui/ca_import_overlay_spec.cr
+++ b/spec/tui/ca_import_overlay_spec.cr
@@ -1,0 +1,118 @@
+require "../spec_helper"
+require "../support/memory_backend"
+require "../support/overlay_harness"
+
+include Gori::Tui
+
+private def ckey(k : Termisu::Input::Key, char : Char? = nil) : Termisu::Event::Key
+  Termisu::Event::Key.new(k, char: char)
+end
+
+# The "Import CA certificate" popup (palette → ca.import). It is a DUMB two-path form:
+# the destructive half — validating the pair, dropping the card, and running the danger
+# confirm before CertAuthority#import! — is the injected on_commit closure at the
+# open-site (Runner#import_ca). These specs pin that split, because the failure mode of
+# getting it wrong is a root CA replaced without a confirm.
+describe Gori::Tui::CAImportOverlay do
+  it "exposes the chrome the collapsed Runner ladders used to hard-code" do
+    OverlayHarness.new(CAImportOverlay.new).assert_chrome(OverlayKind::CaImport, "IMPORT CA")
+  end
+
+  it "walks cert → key with ⇥/↓ and commits only from the LAST row" do
+    ov = CAImportOverlay.new
+    h = OverlayHarness.new(ov)
+    h.on_commit { true }
+
+    # A path under a directory that cannot exist, so the completion dropdown never opens
+    # and ↵ means "next row" rather than "accept a suggestion" (it owns both).
+    h.type("/gori-spec-no-such-dir/ca.pem").should eq(:open)
+    # ↵ on the certificate row advances instead of submitting — a half-filled form must
+    # not reach the confirm.
+    h.press(Termisu::Input::Key::Enter).should eq(:open)
+    h.commits.should eq(0)
+    h.type("/gori-spec-no-such-dir/ca.key").should eq(:open)
+    h.press(Termisu::Input::Key::Enter).should eq(:closed)
+
+    h.commits.should eq(1)
+    ov.cert_path.should eq("/gori-spec-no-such-dir/ca.pem")
+    ov.key_path.should eq("/gori-spec-no-such-dir/ca.key")
+  end
+
+  it "keeps the form up when the commit closure rejects it (a missing path)" do
+    # Runner#submit_ca_import toasts "both … are required" and returns false; the card
+    # must stay so the user can fill the row in, not vanish with the typed path.
+    ov = CAImportOverlay.new
+    h = OverlayHarness.new(ov, commit: false)
+    h.press(Termisu::Input::Key::Down) # → key row
+    h.press(Termisu::Input::Key::Enter).should eq(:open)
+    h.commits.should eq(1) # it ran — it just refused to close
+  end
+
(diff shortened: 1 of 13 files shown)


hahwul:
Coordinator: scope verified — `handle_*_key` 39 → 34 and `Overlay` subclasses 3 → 8, exactly your five modals, `src/gori/tui/overlay.cr` untouched, per-modal spec files following C1's convention.

Extracting `NON_MODAL_KINDS` / `MIGRATED_KINDS` as file-level constants is better than the shape I proposed in my message, which used locals. #371 (C1) independently landed the same derived assertion with local variables, so there is a mechanical conflict between you on those lines. I have asked C1 to keep **your** constants when it rebases, because the remaining three batches each need to append one line to that list and a local would have to be re-declared per example.

Nothing to change here. Merge when CI is green.

PR #371, opened 2026-07-25
refactor(tui): migrate the scan/fuzz config modals onto the Overlay seam (#355)

### LINKED ISSUE #355: refactor(tui): finish the Overlay migration — 28 of 31 modals still ride the Runner ladder

hahwul:
Phase 1 batch **C1** of the overlay-seam migration: the six scan/fuzz config modals move
off the Runner's hand-ordered `case @overlay` ladders onto the polymorphic `Overlay` base.

`notifications`, `probe_active`, `discover_config`, `discover_headers`, `fuzz_set`,
`fuzz_advanced` — each is now an `Overlay` subclass supplying `key`/`title`/`hint`/`render`/
`handle_key` (plus `handle_click`/`move`/`set_preedit` where it had them), and each is deleted
from all seven ladders: IME preedit, key, `MODAL_OVERLAYS`, click, wheel/move, render,
focus-badge title, bottom-row hint. Six `@*_overlay` ivars are gone; the shell holds them in
`@active_overlay`.

`runner.cr` 5,570 → 5,368 lines; `handle_*_key` 39 → 33.

### Domain coupling is injected at the open-site

Every action that belongs to the shell rather than the form is a closure set where the modal
opens, mirroring `ConfirmDialog`:

| Modal | Injected as | Notes |
|---|---|---|
| notifications | `on_commit` (jump to note), `on_palette` (^P hop) | `c` clears the store the overlay already holds, so it stays a plain key case |
| probe_active | `on_commit` → `start_probe_active` | now returns `Bool`; `false` when the selected options would send nothing, so the popup stays up |
| discover_config | `on_commit` → `start_discover`, `on_edit_headers` | `start_discover` returns `Bool`; `false` on the invalid (no spider/bruteforce) path |
| discover_headers | `on_commit` → write back + re-open the parent | see below |
| fuzz_set / fuzz_advanced | `on_commit` → `apply_fuzz_*` | no cancel path: esc, ↵-on-last-field and click-away all apply |

### The `discover_config → discover_headers` nesting

This is trap #1 from the review comment on #355: `close_active_overlay` runs *after* `ov.commit`
and unconditionally clears `@overlay`, so a commit closure that opens another modal gets
clobbered. The headers sub-editor takes the documented workaround — its closure re-opens the
parent itself and returns `false`, so the shell's close-on-commit leaves the restored popup in
place:

```crystal
hov.on_commit = -> {
  cfg.set_headers(hov.headers)
  open_overlay(cfg)
  false
}
```

C5's `property on_close : Proc(Nil)?` is the tree's agreed nesting mechanism and lands after this
batch; folding this call-site onto it is a clean follow-up. This PR does not touch `overlay.cr`.

`^P` out of the notification center has the same ordering hazard in reverse, so its closure
closes first and raises the palette second.

### Click-away semantics are preserved per modal, including the box-nil path

The old ladder was not uniform and the base class's default is `:cancel`, so three of the six
override it. Each was checked against its old `click_*` helper and is now pinned by a spec that
forces the `overlay_box → nil` path with an explicit small `area:` (unreachable through
`OverlayHarness::DEFAULT_AREA`, which is the whole screen):

| Modal | click-away / window-too-small |
|---|---|
| notifications, probe_active, discover_config | dismiss (`:cancel`) |
| discover_headers, fuzz_set, fuzz_advanced | apply (`:commit`) |

### Pre-existing bug found, deliberately NOT fixed

`FuzzAdvancedOverlay#handle_key` discards its `case` value and always returns `:stay`, so
`handle_text`'s commit-on-the-last-row has never reached the shell — ↵ on *Filter regex* is a
no-op, not an apply. The early `return` is still load-bearing as a bounds guard (`@sel` would
otherwise index past `ROWS`). Fixing it is a behaviour change, so this PR documents it at both
sites and pins the *current* behaviour in a spec; it belongs in its own patch. The rendered hint
only promises "esc applies", so nothing advertised is broken.

### `/code-review` findings, all fixed

A correctness pass found no behavioural divergence (it independently re-derived the
`start_discover` ordering analysis and the `offer_flow_headers` capture argument). A test-quality
pass found four specs of mine that passed **vacuously**; each fix was control-run against the
exact mutation that used to survive:

| Spec | Mutation that used to pass | Now |
|---|---|---|
| `fuzz_advanced` click | `ri = @scroll + i` → `ri = i` | fails — `@scroll` only advances inside `render`, so the click example now renders, scrolls, and clicks the scrolled row |
| `notifications` layout.body click | delete `set_selected(idx)` | fails — the example clicked row 0, which is already the default selection |
| `discover_headers` "wheel does nothing" | add a `move` override that scrolls the editor | fails — `headers` re-parses the buffer so it is caret-independent; now compares the whole drawn frame, glyphs and background |
| `discover_headers` / `fuzz_set` / `probe_active` layout.body | hit-test against a hard-coded screen rect | fails — those examples only drove keys, which never see `area`; they now click |

Also replaced the shared gate spec's `unmigrated.size.should eq(28)` literal with derived set
equality (`unmigrated.to_set.should eq(MODAL_OVERLAYS.to_set)`), per the coordination note on
#355. The literal was a same-line conflict for all five batches whose correct value depends on
two independent moving parts; set equality needs no number, leaves each batch appending only to
`migrated`, and is strictly stronger — control-run against an accidental EXTRA member in
`MODAL_OVERLAYS`, which the old count-plus-`includes?` pair could not see.

### Verification

- `just test` → 3966 examples, 0 failures
- `just check` → clean on every touched file; the two `crystal tool format --check` failures
  (`src/gori/proxy/upstream.cr`, `spec/proxy/tls/tunnel_spec.cr`) are pre-existing Crystal-version
  drift, untouched by this branch. ameba per-file counts are identical to `origin/main`
  (`runner.cr` 42 → 42, each overlay unchanged) — no new findings.
- Live TUI smoke test (tmux, isolated `GORI_HOME`): all six modals opened, showed the right focus
  badge and hint, and committed/cancelled correctly — including the `offer_flow_headers` confirm
  round-trip (`return_to: :discover_config` restoring through the `active_overlay` gate), the
  headers sub-editor returning into the Discover popup, `^P` out of the notification center, a
  real active scan firing and toasting, and a fuzz payload set + advanced toggle persisting.

### Note for the other batches

`crystal tool format` re-aligned the `then` column in four ladders (preedit, click, title, hint)
because the widest `when` arm in each (`.discover_headers?`) was deleted. That makes this diff
touch ~90 ladder lines belonging to other batches' modals. It is forced by the formatter, not a
tidy-up. On conflict: resolve on content (keep both sides' deletions), then re-run
`crystal tool format src/gori/tui/runner.cr`.

The shared `spec/tui/overlay_dispatch_spec.cr` edit is held to the two gate examples, which are a
genuine global invariant every batch must update; they are now merged into one derived example so
the only per-batch edit is appending to `migrated`. All new examples live in the per-modal spec
files, which no sibling batch touches.

Closes #355 for batch C1.

### DIFF
--- a/spec/tui/discover_config_overlay_spec.cr
+++ b/spec/tui/discover_config_overlay_spec.cr
@@ -1,5 +1,6 @@
 require "../spec_helper"
 require "../support/memory_backend"
+require "../support/overlay_harness"
 
 include Gori::Tui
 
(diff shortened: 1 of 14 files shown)


hahwul:
Coordinator: scope verified — `handle_*_key` 39 → 33 and `Overlay` subclasses 3 → 9, exactly your six modals, and you correctly leave `src/gori/tui/overlay.cr` alone. Per-modal spec files are the convention the other batches are now following, which came from your PR.

One merge-ordering note. #370 (C2) is also open and also replaced the hard-coded `unmigrated.size.should eq(28)` with the derived set assertion, but with different naming:

- yours: local variables `non_modal` / `migrated`
- C2's: file-level constants `NON_MODAL_KINDS` / `MIGRATED_KINDS`

Same semantics, so this is a mechanical conflict rather than a design disagreement — but whoever merges second should not resolve it by coin flip. **Keep C2's constants.** They are reachable from every example in the file rather than one, which matters because the next three batches each need to append one line to the same list; a local variable would have to be re-declared per example.

Suggested order: #370 lands first, then rebase this PR onto it and fold your six members into `MIGRATED_KINDS`. If you get there first instead, say so on #355 and C2 can adopt your side — the point is that one of you picks, not that it is you.

hahwul:
Rebased onto `d5a11d3` (C2). Two conflicts, both the predicted ladder collisions in `runner.cr` plus the shared gate spec.

**Gate spec:** C2 landed the same derived-set-equality idea with a better shape — a single `MIGRATED_KINDS` ledger constant read by both examples. Adopted theirs verbatim and appended C1's six kinds to the ledger; dropped my duplicate example.

**runner.cr:** the ladder `when` arms conflicted exactly as flagged — both batches deleted from the same alignment-anchored blocks. Resolved deterministically rather than by hand-merging hunks: took the post-C2 file and re-applied C1's deletions to it (they are content-anchored, and C2 never touched these six modals). Every removal was re-verified by count — 3 preedit, 6 key, 6 `MODAL_OVERLAYS`, 6 click, 5 wheel, 6 render, 12 title+hint — plus the six `click_*` helpers, five `handle_*_key` methods and both `apply_close_fuzz_*`.

Re-verified on the new base:
- `just test` → 4002 examples, 0 failures
- ameba on `runner.cr`: 42 findings before, 42 after — none new
- the two mutation controls still fire (`ri = @scroll + i` → `ri = i`; an extra member in `MODAL_OVERLAYS`)
- rebuilt and re-smoked the TUI: DISCOVER → CUSTOM HEADERS → back to DISCOVER, and the notification center's `^P` hop to the palette

Also renamed a local `title` inside `FuzzSetOverlay#render` to `card_title` — it is only a local, so it was never the silent-shadowing bug C2 hit, but the name is now a method on the base and the collision class is worth not re-creating.

hahwul:
Heads up: this landed on `main` red — the merge-commit CI run for `b377a0f` is failing, and `22c35af` before it was green.

```
src/gori/tui/runner.cr:3864: Error: wrong number of arguments for
  'Gori::Tui::Runner#close_active_overlay' (given 0, expected 1)

```

#372 gave `close_active_overlay` a required `ov : Overlay` parameter (runner.cr:1608) and the call site added here passes none, so the two are fine apart and break together. Everything branching off `main` is red until it is fixed — I hit it rebasing #375, which does not touch `runner.cr`.

Not touching it myself since `runner.cr` is the #355 agents' file.

PR #372, opened 2026-07-25
refactor(tui): give Overlay a nested-modal seam (on_close) (#355)

hahwul:
Phase 1 of #355 migrates 28 modals onto the `Overlay` seam in five parallel batches, and two of those batches need a modal to open **another** modal and be returned to when it closes. The shell can express exactly one such relationship today — a `@prefs_return` flag plus a `settle_sub_editor` call at each dispatch chokepoint — usable only by the Preferences family that owns it.

Split out of the **C5** batch and landed on its own so **C3** (`confirm`) can build on it instead of inventing a second mechanism in the same file. No modal migrations here.

## The seam

`Overlay#on_close : Proc(Nil)?` — what the shell runs after this overlay closes, on a commit or a cancel. A modal opened from another supplies `-> { open_overlay(parent) }`, so ↵-ing into the Theme editor from Preferences and pressing esc lands back in Preferences rather than on the bare tab body.

A proc rather than a `return_to : Overlay?`, because the restore is not always just "re-open the parent": returning from the Hostnames editor also has to re-pull the Preferences modal's Network section, whose "N entries" row that editor just moved. It also composes, which one shell-wide flag could not — a modal can nest inside a modal that is itself nested.

Two ordering properties are load-bearing and both are spec'd:

- `close_active_overlay` drops the modal **first** and runs `on_close` after, so a pop-back's `open_overlay` is the last write.
- `leave_overlay` is the deliberate escape hatch that **skips** it, for an exit going somewhere else entirely (^P to the palette), where a pop-back would re-open on top of the destination.

`close_active_overlay` also now takes the overlay it is closing and checks it against what the shell currently holds. A handler may change that mid-key (a Preferences opener row opens its editor; ^P calls `leave_overlay`) and both must return `:stay` — re-reading `@active_overlay` meant a wrong return would drop the editor that had just appeared and run *its* pop-back, or blank the palette ^P had just opened. Latent rather than live today (the three migrated modals are leaf forms), but the seam exists for the batches about to add exactly those handlers.

**No behaviour change**: nothing sets `on_close` yet, so every close runs a nil proc.

## For C3

The three sites where a confirm is raised from inside another modal are covered, and each has an example:

| site | parent | restored as |
|---|---|---|
| settings ^R reset | `SettingsOverlay` (migrated) | the object, via `open_overlay` |
| tab-bar reset | `TabsOverlay` (migrated) | the object, via `open_overlay` |
| `history_controller.cr:385` delete-from-detail | `:detail` (**not** migrated) | the `@overlay` state |

Cancel lands back in the parent for all three.

Two things `confirm()` has to get right, both spec'd here:

1. **Capture the parent before `open_overlay`.** Today `confirm()` never touches `@active_overlay`, so the parent goes inert but stays held. Once ConfirmDialog is opened through `open_overlay`, that reference is overwritten.
2. **Run the action from `on_close`, after the restore — not from `on_commit`.** `run_confirm` is restore-then-action today, and `history_controller.cr:385`'s action reads `@host.overlay == :detail` to decide whether the drill-in should close now the flow is gone. The generic dispatch runs `commit` *before* the close, so an action wired into `on_commit` reads the dialog's own state, the guard never fires, and the pop-back then re-opens the detail on a deleted flow. Two examples state both halves.

## Verification

- `just test` → **3926 examples, 0 failures**
- `crystal tool format --check` clean on the touched files; `ameba` byte-identical to the `origin/main` baseline for the same files (zero new findings)
- Every example control-run against the mutation that motivated it: removing the harness's `on_close` call, inverting the drop/close order, closing on a REJECTED commit, and dropping the identity check each fail the file. The first of those initially did **not** fail — the examples asserted `harness.closes` (a counter that increments however the harness behaves) instead of the overlay's own closure; they now count the closure.

`OverlayHarness` learns `on_close` because it is the spec-side statement of `dispatch_overlay_key` / `close_active_overlay`; a change to that dispatch has to land there too, or every downstream migration spec passes against a stale double.

## Also

`spec/tui/overlay_dispatch_spec.cr`'s input-capture gate is now a **derived set** rather than a hard-coded total of 28. All five Phase 1 batches edit that assertion, and five agents editing one integer to five different values merges badly, whereas five appending to a one-member-per-line list does not. Both directions still fail on mutation: a migrated kind left in `MODAL_OVERLAYS`, and an unmigrated kind deleted from it.

Refs #355

### DIFF
--- a/spec/support/overlay_harness.cr
+++ b/spec/support/overlay_harness.cr
@@ -15,6 +15,8 @@ require "./memory_backend"
 #   * `open?` mirrors whether the shell still holds the modal in `@active_overlay`.
 #   * `commits` counts runs of the injected `on_commit` closure — the open-site behaviour
 #     a migrated modal no longer carries itself.
+#   * `closes` counts runs of `on_close`, the nested-modal pop-back the shell performs
+#     AFTER dropping the modal (see the ordering note on `close!`).
 #
 # If those Runner methods ever change, change them HERE too: this harness is the
 # spec-side statement of that contract, and a modal spec that passes against a stale
@@ -46,6 +48,9 @@ class OverlayHarness
   getter area : Gori::Tui::Rect
   # Times the injected on_commit ran (a :commit outcome, whether or not it closed).
   getter commits = 0
+  # Times the overlay's on_close ran — the nested-modal pop-back, which fires on a cancel
+  # AND on a commit that closed, but never while the modal stays up.
+  getter closes = 0
   # The shell still holds this modal — false once a :cancel, or a :commit whose closure
   # returned true, dropped it.
   getter? open = true
@@ -176,10 +181,20 @@ class OverlayHarness
 
   private def dispatch(outcome : Symbol) : Symbol
     case outcome
-    when :cancel then @open = false
-    when :commit then @open = false if @overlay.commit
+    when :cancel then close!
+    when :commit then close! if @overlay.commit
     end
     @open ? :open : :closed
+  end
+
+  # Mirrors `Runner#close_active_overlay`: the shell DROPS the modal first and runs its
+  # `on_close` after. That order is load-bearing rather than incidental — a pop-back
+  # closure re-opens its parent with `open_overlay`, so it has to be the last write;
+  # running on_close first would leave the shell holding nothing.
+  private def close! : Nil
+    @open = false
+    @closes += 1
+    @overlay.on_close.try(&.call)
   end
 
   private def event(k : Termisu::Input::Key, char : Char?,
--- a/spec/tui/overlay_nesting_spec.cr
+++ b/spec/tui/overlay_nesting_spec.cr
@@ -1,0 +1,350 @@
+require "../spec_helper"
+require "../support/memory_backend"
+require "../support/overlay_harness"
+
+include Gori::Tui
+
+# `Overlay#on_close` — the nested-modal seam (#355). A modal opened FROM another pops back
+# into it on close, which the shell used to express with a single `@prefs_return` flag plus
+# a `settle_sub_editor` call at each dispatch chokepoint: one hard-coded relationship for
+# the whole app, and one the Preferences family alone could use.
+#
+# Everything the migration batches need from it is a CONTRACT about shell state, not about
+# any one modal, so this file pins the contract with doubles and no production overlay:
+#
+#   * on_close runs on a cancel and on a commit that closed — never while the modal is up
+#   * the shell drops the modal BEFORE running it, so a pop-back's open_overlay sticks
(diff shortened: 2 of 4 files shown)


hahwul:
Rebased onto `d5a11d3` (C2 / #370). Two changes from what was first pushed:

**Dropped my `spec/tui/overlay_dispatch_spec.cr` change.** C2 landed the same
count-to-derived-set fix and did it better — `MIGRATED_KINDS` / `NON_MODAL_KINDS` hoisted
to file constants that both examples read, so a batch edits one list and nothing else. I
took theirs wholesale; that file is now byte-identical to main and out of this diff.

**`submit_ca_import` picked up the new signature.** C2 added a `close_active_overlay` call
inside a commit closure, and its comment describes precisely the hazard the identity check
in this PR removes:

> The rule the others follow — return true and let the shell close you — does not work
> here: the shell's post-commit close unconditionally resets `@overlay`, which would wipe
> the confirm this opens one line after opening it. So it closes first and returns false,
> meaning "already handled, keep your hands off `@overlay`".

That `return false` is still the right call and is untouched. But the "keep your hands off
`@overlay`" part is now enforced by the shell rather than by the closure's return value: a
second `close_active_overlay(ov)` finds `@active_overlay` already empty and no-ops,
instead of blanking `@overlay` out from under the confirm.

`just test` → **3962 examples, 0 failures** on the rebase. Format and ameba clean.

hahwul:
Coordinator: this is exactly the split I asked for, and it answered the condition I set rather than deferring it. Verified — 4 files, +420/-8, no modal migrations, `Overlay` subclass count unchanged.

The three ConfirmDialog sites are mapped with the not-yet-migrated parent (`:detail`) handled through the `@overlay` state rather than an object, which is the case I asked you to check and could not have been answered by hand-waving.

Two things here are worth more than the seam itself:

**The `on_close` vs `on_commit` ordering trap.** `history_controller.cr:385`'s action reading `@host.overlay == :detail` while the generic dispatch runs `commit` before the close means an action wired into `on_commit` silently re-opens a detail view on a deleted flow. C3 would have hit that and spent real time on it. Finding it in the base PR rather than in C3's migration is the whole return on splitting.

**Disclosing that one example did not fail under mutation on the first try** — asserting `harness.closes` (a counter that increments however the harness behaves) instead of the overlay's own closure. That is the same class of defect as #369, and it is only visible because you control-ran every example against the mutation that motivated it. A green suite would have hidden it.

I have relayed both traps to C3 and told it to rebase onto this branch rather than wait, and not to add `on_close` itself.

The `@prefs_return` + `settle_sub_editor` removal argument holds: one shell-wide flag could express exactly one nesting relationship, and a proc composes. Merge when you are ready — C3 is waiting on this.

Issue #367, opened 2026-07-25
TUI workbench tools ignore hostname overrides (Fuzzer/Miner/Sequencer/Repeater/Discover)


hahwul:
Surfaced while reviewing #356 (the Fuzz Plan builder), which noted it for the Fuzzer tab and added an `overrides` parameter defaulting to `nil` so the fix is a one-call-site edit. Checking the rest of the tree, the gap is **not** Fuzzer-specific.

## The asymmetry

Every active sender takes `overrides : Gori::HostOverrides? = nil`. Every CLI and MCP tool passes it. **No TUI tool does**, so they all silently take the `nil` default:

| Surface | Passes host overrides? |
|---|---|
| `cli/run/fuzz.cr:117` | yes — `cli_host_overrides(project_name, db_path, flow_id)` |
| `cli/run/mine.cr`, `cli/run/sequence.cr` | yes |
| `mcp/tools/fuzz.cr:245` | yes — `HostOverrides.load(store)` |
| `mcp/tools/mine.cr`, `mcp/tools/sequence.cr` | yes |
| `tui/fuzzer_view.cr:943` | **no** |
| `tui/miner_view.cr` | **no** |
| `tui/sequencer_view.cr` | **no** |
| `tui/controllers/repeater_controller.cr:1409` | **no** |
| `tui/controllers/discover_controller.cr` | **no** |

For example, on `origin/main`:

```crystal
// src/gori/tui/fuzzer_view.cr:943 — no `overrides:` argument
sender = Fuzz::Sender.new(Fuzz::Origin.new(scheme, host, port), Gori::Outbound.interactive(scope),
  http2: @http2, verify: verify, sni: sni_override, timeout: @config.timeout)
```

## Why it matters

The proxy path does honour overrides (`Proxy::Upstream.dial`), so **captured** traffic is fine. The workbench tools dial directly and bypass that path entirely.

The failure mode is quiet and points the wrong way. An operator pins `api.prod.example.com` to a staging IP in the Project tab's HOST OVERRIDES pane, then runs the Fuzzer against that host from the TUI. The override is ignored, DNS resolves for real, and the payloads land on production. The same run through `gori run fuzz` would have been pinned correctly. Nothing in the UI indicates the difference.

Note this is orthogonal to scope: #354 made the scope gate uniform, and scope would still allow the host here — the target is in scope, it is just the wrong machine.

## Scope of the fix

[...]

PR #375, opened 2026-07-25
refactor(repeater): one Plan builder for TUI/CLI/MCP send assembly (#356) (#367)

hahwul:
The repeater slice of #356's fan-out, plus the repeater slice of #367.

`Repeater::Plan.build(options, outbound)` is now the only place a `Repeater::Sender` is
constructed (`grep -rn 'Repeater::Sender.new' src/` matches `repeater/plan.cr` alone). The
sequence *expand → Content-Length policy → WebSocket detection → target parse → SNI → host
overrides → sender* used to exist five times over — `gori run repeater` for a flow and for a
saved session, MCP `send_request` and `send_websocket`, and the TUI's ^R / send-group / WS
paths — and the copies had drifted.

## Behaviour I unified, and which answer I picked

| Drift | Was | Now |
|---|---|---|
| Host overrides | every CLI/MCP tool passed `overrides:`, **no** TUI tool did (#367) | the builder threads it; the TUI passes the session's **live** `Session#host_overrides` |
| `Env.expand` over SNI | MCP's flow path sent a captured `$SNI_HOST` literally; the CLI expanded it | expanded once, in the builder, on every surface |
| `port <= 0` | rejected only by MCP `send_websocket`; the other four dialled it | rejected everywhere (`BadTarget`) |
| Content-Length | `--no-auto-cl` and `-H "Content-Length: N"` were two open-coded spellings of one knob | one `auto_content_length` knob; `build_single_flow_request` returns the pre-expansion wire + `explicit_cl` |

For #367 I pass the session's **live** `HostOverrides` instance rather than a fresh
`HostOverrides.load(store)`: the proxy reads that one and the Project tab's HOST OVERRIDES
pane edits it under a Mutex, so a copy taken at send time would miss every edit made after
the tab opened.

**Deliberately not unified:** `expand_request: false` for the TUI editor's hex / gRPC /
decode / §…§ modes and for MCP's `RequestBuilder`, all of which already produce final bytes.
A second `Env.expand_wire` would double-expand a `$KEY` whose value itself looks like a token.

## A TLS downgrade the refactor surfaced

`WsEngine` decides TLS with `scheme == "https" || scheme == "wss"`, but `Engine` and
`H2Engine` test `== "https"` **alone**. A WebSocket upgrade replayed as a plain one-shot
therefore went out over a **cleartext** socket to a TLS port, leaking the request's cookies
and auth headers. Reachable as:

```
gori run repeater <flow-id> --target wss://host
```

for a captured upgrade whose response was **not** 101, so the command's 101 guard never
fires. The builder now folds `ws`→`http` and `wss`→`https`, which makes every engine's TLS
test correct by construction and closes the same latent hole on the TUI and MCP paths.
Regression test drives a raw TCP listener and asserts the secret never reaches it.

## Errors

`PlanError::Reason` (`NoRequest` / `NoTarget` / `BadTarget` / `UnsupportedScheme`), matched
exhaustively with `in` at each surface so each words the refusal in its own vocabulary — the
TUI names its TARGET pane, the CLI its flags, MCP a coded `INVALID_ARGUMENT` field.

One intentional divergence from `Fuzz::PlanOptions`: a **blank** explicit target is an error,
not "fall back to the default". No repeater surface sends `"url": ""` (MCP only ever supplies
`default_target`/`origin`), and `gori run repeater --target "$UNSET"` must not quietly
redirect the send to the captured host.

## Also in here

The three minimize call sites (CLI / TUI / MCP) dial `Fuzz::Sender` directly rather than
through `Plan`, so they get `overrides:` threaded by hand; the TUI one also gets `Env.expand`
over its SNI, which the other two always had. Without this, MCP minimize would have been the
one repeater send path still ignoring the operator's pin.

## Verification

- `crystal spec spec/repeater spec/cli_run_spec.cr` → 166 examples, 0 failures. Full suite green.
- **Mutation-tested, not just green:** disabling the builder's CL resync kills 6 examples;
  changing `auto_content_length: rec.auto_content_length?` to `true` fails the restored CLI
  mapping spec; dropping `overrides:` fails the host-override spec. `/code-review` caught that
  my first draft of the CLI spec asserted `Plan` against `Plan` and never executed the CLI's
  row→options mapping at all — extracted `session_plan_options` and pinned it.
- **End-to-end against a real origin:** `repeater send` with auto-CL on (`cl=5`) and off
  (`cl=2`, origin read a truncated body — the deliberately-wrong CL survived); single-flow
  replay with `-b` (`cl=7`) and with `-H "Content-Length: 3"` (`cl=3`); `--target` override.
- **Real TUI via tmux:** ^R and send-group against a host that is **NXDOMAIN**
  (`pinned.invalid`, where `curl` fails with exit 6) both reached the origin through the
  project's host override — the #367 fix working in the actual app, not just in a spec.
- `wss://` on the CLI now fails the TLS handshake against a plaintext origin instead of
  leaking the request.

## Not closing the issues

Marked "Part of" rather than "Closes": #356 and #367 each cover four other tools still in
flight, and a sibling PR that used `Closes` auto-closed both and they had to be reopened.

`DESIGN.md` §7 is deliberately untouched — four agents appending to the same section would
conflict; worth one entry covering all four builders once they land.

## ⚠️ Base is red, not from this PR

`origin/main` at `b377a0f` does not compile:

```
src/gori/tui/runner.cr:3864: Error: wrong number of arguments for
  'Gori::Tui::Runner#close_active_overlay' (given 0, expected 1)
```

`close_active_overlay` gained a required `ov : Overlay` parameter in #372 (`22c35af`) and the
call site added by #371 (`b377a0f`) passes none. This branch does not touch `runner.cr`
(`git diff --quiet origin/main HEAD -- src/gori/tui/runner.cr` is clean), so CI here will be
red until main is fixed. Holding the merge on that.


hahwul:
**Blocked on a red base, not on this change.**

CI here fails in `build-crystal` before any of this PR's code is reached:

```
src/gori/tui/runner.cr:3864
Overloads are:
 - Gori::Tui::Runner#close_active_overlay(ov : Overlay)
```

The `tests` jobs then show as failed only because fail-fast cancelled them — their logs end at `shards install` with `The operation was canceled`, so no spec of mine ran either way.

Evidence this is inherited:

- `main`'s own CI is failing at `b377a0f` and was green at `22c35af` immediately before it.
- `git diff --quiet origin/main HEAD -- src/gori/tui/runner.cr` is clean — this branch does not touch that file.
- #372 gave `close_active_overlay` a required `ov : Overlay` parameter (runner.cr:1608); the call site added by #371 passes none.

Flagged on #371. Holding the merge until `main` compiles, then I will rebase and re-run.

Local verification in the meantime: `crystal spec spec/repeater spec/cli_run_spec.cr` → 166 examples, 0 failures; full suite green; the CL-resync, row→options and host-override specs each verified by mutation; and both the CLI and TUI paths exercised end-to-end against a live origin.

PR #377, opened 2026-07-25
refactor(sequencer): one Plan builder for TUI/CLI/MCP run assembly (#356, #367)

### LINKED ISSUE #356: refactor(tools): extract per-tool Plan builders so TUI/CLI/MCP share run assembly

### LINKED ISSUE #367: TUI workbench tools ignore hostname overrides (Fuzzer/Miner/Sequencer/Repeater/Discover)

hahwul:
Closes #356 (sequencer slice). Closes #367 (sequencer slice).

The sequencer's "assemble a run" pipeline existed three times over — `SequencerView#build_engine`, `cli/run/sequence.cr`, and MCP `build_sequence_job` — and the copies had drifted. This does for the sequencer what #366 (`7dc9908`) did for the fuzzer, following that builder as the convention.

`Sequencer::PlanOptions` (a plain struct) plus `Sequencer::Plan.build(options, outbound)` splits the two jobs that were tangled: parsing an input format stays surface-specific (OptionParser on the CLI, the args hash on MCP, view state in the TUI), while everything downstream of the normalized options has one implementation. `Sequencer::Engine.new` now has exactly two call sites, both inside `plan.cr`.

The `Outbound` stays an **argument**. Layer-1 strictness differs per surface on purpose (`Outbound.agent` / `.cli` / `.interactive`, DESIGN.md §7); building one inside the builder would collapse that distinction.

## Manual mode no longer needs a fake sender

Manual mode analyses a pasted token list and sends nothing, but the engine's constructor demanded a backend, so the TUI handed it:

```crystal
dummy = Fuzz::Sender.new(Fuzz::Origin.new("http", "localhost", 80), outbound, http2: false, verify: false)
```

a real dial target for a run that must never dial (and #354 then had to give that dummy an `Outbound` too). Modelled honestly instead: `Sequencer::Engine`'s backend is `Fuzz::Backend?` and an analyse-only `Plan` carries no sender, no origin and no request. **This is the Engine signature change flagged in the task brief** — live replay without a backend is rejected in the constructor, and a consumer that flips the shared `Config` to `LiveReplay` after the fact gets a clean `ErrorEvent` and no traffic (specced, `plan_spec.cr`).

## Divergences the builder settles

- **`Env.expand` runs once**, on the request bytes and on the resolved target. The TUI ran it on **neither**, so a `$TOKEN` in a sequenced request went out literally there while resolving on the other two surfaces; MCP expanded a flow's target **twice** (`mine_request_source` + `fuzz_origin`), and a second pass resolves tokens the first pass produced. The TUI's seed path (`text_to_request`) stopped expanding as well, so a pasted request keeps its `$TOKEN`s in the persisted session the way the Repeater editor does.
- **The TUI applies the project's HOST OVERRIDES (#367).** It dialled real DNS while `gori run sequence` pinned the override, so the same request from two surfaces could land on two different machines with nothing in the UI to say so. Per #367 the controller passes the session's **live** `Session#host_overrides` (the instance the pane edits and the proxy reads, Mutex-guarded), not a second `HostOverrides.load(store)` copy that would miss later edits. The parameter has **no default**, so a call site cannot inherit the bug back.
- **One blank-descriptor refusal** in the builder instead of three (the TUI's inline check, the CLI's abort in `build_token_loc`, MCP's exactly-one-of rule).

Each surface keeps its own error vocabulary: `Sequencer::PlanError` carries a machine-readable `reason` and each surface words it in its own idiom (`--cookie/--header/...` on the CLI, `cookie|header|regex|position|jsonpath` on MCP, the config overlay in the TUI).

Also: `gori run sequence` now hydrates the project's env vars explicitly for a `--request`/stdin source, the way `cmd_fuzz` already does, instead of depending on the store `cli_host_overrides` happens to open.

## Verification

All three surfaces driven live (`just build` first — `crystal spec` does not rebuild `./bin/gori`):

- **CLI** — live replay against a local token-issuing server (cookie + regex descriptors), every refusal path (no target / blank selector / bad target / bad scheme / no descriptor / bad `--position`) with messages unchanged from `HEAD`, `--tokens`, and a **project** env var resolving in `--target` on the `--request` path.
- **MCP** over stdio JSON-RPC — `sequence_start` → `sequence_status` → `sequence_results` (8 tokens collected), all five refusals, and the scope gate (`no scope is configured … pass allow_unscoped:true`).
- **TUI** under tmux — added a HOST OVERRIDE for `nonexistent.invalid` (no DNS record anywhere), captured a flow through the proxy, sent it to the Sequencer: **500/500 collected**. Deleted the override mid-session and re-ran: **0 collected, every sample `connect failed: … host unreachable`**. That is #367 end to end, live, in both directions. Manual paste verified separately: `1/1 collected · 0 sent`.
- **Specs** — new `spec/sequencer/plan_spec.cr` (table-driven: CLI-, MCP- and TUI-shaped options build a byte-identical run, pinned against literals first so three surfaces agreeing on a wrong plan cannot pass silently) and new `spec/tui/sequencer_view_spec.cr` (the TUI mapping itself, incl. a loopback A/B where `overrides` is the only variable). `just test` green.

`/code-review` at deep depth ran three specialists plus an adversarial + spec-quality pass; it mutation-tested the suite and found five mutations that slipped through silently, including **#367 reintroduced at the TUI call site**. Fixed by pinning `http2`/`token_loc`/`concurrency` against literals, adding a config-identity spec, adding the analyse-only-cannot-send spec, rewriting the override control as a same-port A/B with a non-vacuous assertion, and adding the view-level spec. All five now fail on their mutation (re-verified by hand).

## Known behaviour changes, called out deliberately

1. A blank `--target=` on the CLI now means "use the seeding flow's target" instead of aborting (`.presence`), matching `Fuzz::Plan` and MCP's contract.
2. MCP reports a missing token location *before* a missing `url` when both are absent (it validates the descriptor while parsing). Message-only; both are `isError`.
3. `Env.expand_wire` normalizes a bare-LF request **head** to CRLF, so a TUI-sequenced flow with LF framing is now CRLF-normalized — as `gori run sequence`, MCP, and the TUI Repeater have always done.
4. On the CLI, the "no project in play … sending UNSCOPED" notice now prints before a validation abort, since the `Outbound` is built before `Plan.build`. Same as `gori run fuzz` post-#366.

## Not fixed here (out of lane / needs its own decision)

- `SequencerView#target_origin` does not `Env.expand`, so a `$VAR` target displays and persists unexpanded while the dial resolves it. `FuzzerView` expands there; unifying also changes what `insert_sequencer_session` stores, so it wants its own issue.
- A `Position` descriptor with `pos_end <= pos_start` (including the `0:0` default) can extract nothing yet still spend `max_sends` live requests. Pre-existing; a proper refusal needs a fifth `PlanError::Reason` and three surface sentences.
- Scheme validation stays CLI-only (MCP and the TUI never had it), exactly as `Fuzz::Plan` left it.
- No DESIGN.md entry: #366's §7 entry already states the pattern and names Sequencer as intended, and four Plan-builder PRs appending to one file in parallel would collide.

## ⚠️ `main` is currently broken, before this branch

`origin/main` at b377a0f (#371) does not compile:

```
In src/gori/tui/runner.cr:3864:28
```

#372 (`22c35af`) changed `close_active_overlay` to take an `Overlay`; #371 merged on top with a zero-arg call site. Verified by building `origin/main` alone in a detached worktree — nothing to do with this branch, and `runner.cr` is owned by the #355 agents so I have not touched it. This commit cherry-picked onto `22c35af` (the last green commit) is **4025 examples, 0 failures**. CI here will stay red until `main` is fixed.


hahwul:
CI is red on an inherited breakage, not on this branch.

Every failing job dies at the same single error, which is in `main` itself (`b377a0f`, #371):

```
In src/gori/tui/runner.cr:3864:28
```

#372 (`22c35af`) gave `close_active_overlay` a required `Overlay` argument; #371 merged on top with the old zero-arg call site. Confirmed by building `origin/main` on its own in a detached worktree — same error, no commit of mine present. Nothing in the failure logs comes from this diff, and `runner.cr` belongs to the #355 agents so I have not touched it.

**#376 is the fix** (same line, `leave_overlay` instead). This commit cherry-picked onto `22c35af` — the last commit that compiles — is 4025 examples, 0 failures locally.

Holding this PR unmerged. Once #376 lands I will rebase and re-run CI; no changes to the sequencer work are needed for it.

hahwul:
Rebased onto `439ce2f` — #376 landed the `leave_overlay` fix, so `main` compiles again and this branch is no longer inheriting the break.

Locally on the rebased branch: `crystal build --no-codegen src/main.cr` clean, `just test` 4079 examples / 0 failures. Re-running CI.

(#378 proposes the same one-line fix and is now redundant against current main.)

PR #378, opened 2026-07-25
fix(tui): main does not compile — close_active_overlay arity after #371 + #372

hahwul:
`origin/main` (`b377a0f`) does not compile. #371 and #372 were each CI-green on their own branch and broke the tree when combined.

```
In src/gori/tui/runner.cr:3864:28
```

#372 split the teardown into `close_active_overlay(ov)` (drops the modal **and** runs its `on_close`) and `leave_overlay` (drops it **without**). #371 added the call site above against the old zero-arg signature.

**The fix is `leave_overlay`, not `close_active_overlay(ov)`.** #372's comment names this call site as the reason `leave_overlay` exists — the ^P jump to the palette, "where a nested modal's pop-back would otherwise re-open on top of the destination". Passing `ov` would compile and then re-open the notification centre over the palette.

Verified locally: `crystal build --no-codegen` clean; `overlay_dispatch_spec` + `overlay_nesting_spec` 42 examples, 0 failures.

Blocking C3, C4, C5 (#355) and plan-sequencer, plan-repeater (#356) — #375 and #377 are both red for this reason and not for anything in their own diffs.

Refs #355

### DIFF
--- a/src/gori/tui/runner.cr
+++ b/src/gori/tui/runner.cr
@@ -3861,7 +3861,7 @@ module Gori::Tui
       }
       # Close BEFORE raising the palette: the reverse order would drop @active_overlay on
       # top of the modal we just opened.
-      ov.on_palette = -> { close_active_overlay; open_palette }
+      ov.on_palette = -> { leave_overlay; open_palette }
       open_overlay(ov)
       @notifications.mark_all_read
     end


hahwul:
Confirming this is the correct fix, and that the break is mine — #372 changed the signature.

`leave_overlay` is right, not `close_active_overlay(ov)`. This call site is literally the reason `leave_overlay` exists: `close_active_overlay` runs the modal's `on_close`, and for a nested modal that pop-back would re-open the parent **on top of** the palette the user just asked for. `leave_overlay` drops the modal and stops there.

The existing comment above the line ("Close BEFORE raising the palette: the reverse order would drop `@active_overlay` on top of the modal we just opened") still reads correctly with `leave_overlay` — the ordering point it makes is unchanged.

On how it got in: this was a semantic conflict, not a textual one. #371 added the call site and #372 changed the arity, in different regions of `runner.cr`, so git merged them cleanly and each branch was green alone. Worth noting the compiler *did* catch it the moment the two met — the failure mode was CI never building the merge result, not a silent runtime break.

One follow-up rather than a change here: C5 (#379) adds a `jump_to_palette` helper that is exactly this `leave_overlay; open_palette` pair, because five migrated modals needed it. Once both land, this line and those five can collapse onto it. Not worth touching in a red-main fix.

hahwul:
Superseded by #376, which landed the identical fix as `439ce2f` before this was opened.

#376 reached the same conclusion I did — `leave_overlay`, not `close_active_overlay(ov)` — and documented it better, also noting that the ordering matters in the other direction (closing after `open_palette` would drop the palette).

I opened this without seeing #376 because it was on a branch outside my watch list and opened and merged between two of my polling ticks. My duplicate, not a disagreement. Closing unmerged; `origin/main` compiles again at `439ce2f`.

PR #379, opened 2026-07-25
refactor(tui): migrate the Preferences family onto the Overlay seam (#355)

### LINKED ISSUE #355: refactor(tui): finish the Overlay migration — 28 of 31 modals still ride the Runner ladder

hahwul:
Batch **C5** of #355 — the unified Preferences modal and the five dedicated editors it opens: the theme card, the tab-bar customizer, the global hostname overrides, the environment variables, and the hotkey rebinder. Each was a `handle_*_key` method plus an arm in all seven `case @overlay` ladders (preedit, key, click, wheel, render, title, hint). Each is now an `Overlay`, with its Runner coupling injected as closures at the open-site.

Builds on #372 (the `on_close` seam), which was split out of this batch and merged first.

| | before | after |
|---|---|---|
| `runner.cr` lines | 5,585 | **5,062** |
| `handle_*_key` in `runner.cr` | 31 | **25** |
| `runner.cr` references to these six `OverlayKind`s | 60 | **0** |

## The nested lifecycle

This batch owned `@prefs_return` + `settle_sub_editor` — a shell-wide flag plus a call at each dispatch chokepoint, whose only job was landing a sub-editor back in the Preferences modal it was opened from. Both are gone. Each editor's open-site supplies its own `on_close`, so the relationship is per-overlay and composes rather than being one bit for the whole shell.

`^P` goes through `jump_to_palette` (`leave_overlay` + `open_palette`), which deliberately skips `on_close` so the pop-back cannot re-open Preferences on top of the palette — the case the old flag handled with `@overlay.none?` inside `settle_sub_editor`.

The two resets that raise a confirm from *inside* an editor (settings ^R, tab-bar `r`) still use `return_to:`. Raising a confirm only moves `@overlay`, so `@active_overlay` keeps holding the editor and the existing liveness gate brings it back when the confirm closes.

## The capture-mode carve-out

The global ^C/^D quit-arm must be skipped while the hotkey rebinder is recording a chord, so those chords can be **bound** (reserved.cr rejects them inline with "Ctrl-C/D quits gori" while staying in capture) instead of arming a quit. That check read `@overlay.hotkeys? && @hotkeys_overlay.capturing?`; with the ivar gone it asks the modal through `Overlay#raw_key_capture?`, default false. Precedence, `@quit_armed` and `@toast` handling are unchanged.

## Shape

The four Runner-only editors subclass `Overlay` directly. `SettingsView` and `PreferencesView` do **not**: three surfaces share them — the Preferences modal's inline sections (`render_fields_into`), the project picker, and these cards — and only one is a Runner modal. So the modal is a thin `SettingsOverlay` / `PreferencesOverlay` around the unchanged form engine, which also leaves `project_picker.cr` untouched.

The add/edit sub-modes of the hostnames and env editors (and env's prefix editor) fold into their parent overlay, so esc there cancels the **sub-mode** and leaves the modal up, exactly as the three separate Runner handlers did. `env_overlay_hints` becomes `EnvOverlay#hint`.

## Verification

`just test` → **4,010 examples, 0 failures**. `spec/tui/overlay_c5_spec.cr` adds 48 examples driving all six through `OverlayHarness`.

Also smoke-tested against the real TUI, because the specs reach the shell only through doubles:

- palette → Settings → PREFERENCES → Appearance → ↵ on the Theme opener → the `SETTINGS · THEME` card; ↓ live-previews; **esc lands back in PREFERENCES** with the preview reverted and nothing persisted
- PREFERENCES → Editor & Keys → ↵ on Hotkeys → `e` → **^C is rejected inline with "Ctrl-C quits gori", stays in capture, and gori does not quit**
- esc unwinds one level at a time: capture → HOTKEYS → PREFERENCES → TABS
- palette → `Settings: Tabs` → `r` → CONFIRM → esc **lands back in TAB BAR**

Two spec bugs of my own were caught by running them rather than by reading: an off-by-three backspace count, and a sequence that overshot the add row, cancelled it, and then had `e` land in the *list* as edit-start — where the entry legitimately is not a duplicate.

`ameba` on the touched files: 52 findings vs **61** on the same files at `main` — no new categories, no new instances.

Closes #355

### DIFF
--- a/spec/tui/overlay_c5_spec.cr
+++ b/spec/tui/overlay_c5_spec.cr
@@ -1,0 +1,788 @@
+require "../spec_helper"
+require "../support/memory_backend"
+require "../support/overlay_harness"
+
+include Gori::Tui
+
+# Batch C5 of the Overlay migration (#355): the Preferences family — the unified modal and
+# the theme / tab-bar / hostnames / env / hotkeys editors it opens.
+#
+# Everything here rides OverlayHarness, which replays Runner#dispatch_overlay_key /
+# #dispatch_overlay_click, so these lock the routing runner.cr no longer spells out:
+# a handle_*_key method and six `case @overlay` arms per modal were deleted, and the
+# behaviour they encoded now has to live in the overlay or in an injected closure.
+#
+# The three things worth being careful about, and why each gets examples below:
+#   * the ADD/PREFIX sub-modes of the hostnames + env editors folded into their parent
+#     overlay, so esc there must cancel the SUB-MODE and leave the modal up;
+#   * the hotkey rebinder's capture mode, which the shell routes to ahead of its own
+#     ^C/^D quit-arm — the overlay answers for that through `raw_key_capture?`;
+#   * the nested lifecycle, which was a shell-wide `@prefs_return` flag and is now
+#     `Overlay#on_close`, running on a cancel AND on a commit that closed.
+#
+# CAVEAT on driving keys: these overlays branch on `ev.key` for navigation (`key.lower_k?`)
+# and on `ev.char` for mnemonics, exactly as the deleted Runner handlers did. So a mnemonic
+# is pressed as `press(Key::LowerA, 'a')` and vim-nav as `press(Key::LowerK)` — `type` would
+# send every char on Key::LowerA and silently miss the nav branch.
+
+private ESC   = Termisu::Input::Key::Escape
+private ENTER = Termisu::Input::Key::Enter
+private SPACE = Termisu::Input::Key::Space
+private LEFT  = Termisu::Input::Key::Left
+private RIGHT = Termisu::Input::Key::Right
+private DOWN  = Termisu::Input::Key::Down
+private UP    = Termisu::Input::Key::Up
+private TAB   = Termisu::Input::Key::Tab
+private BKSP  = Termisu::Input::Key::Backspace
+private ALPHA = Termisu::Input::Key::LowerA
+
+# Press a printable mnemonic the way the terminal delivers it: the real char attached, so
+# `ev.char` sees it. The key itself is irrelevant to every mnemonic branch in this family.
+private def mnemonic(h : OverlayHarness, c : Char) : Symbol
+  h.press(ALPHA, c)
+end
+
+# ^P — the one chord every modal in this family claims, to leave for the command palette.
+private def ctrl_p(h : OverlayHarness) : Symbol
+  h.press(Termisu::Input::Key::LowerP, ctrl: true)
+end
+
+# ---------------------------------------------------------------------------------------
+# TAB BAR
+# ---------------------------------------------------------------------------------------
+
+# Hide every tab but one, so the next space hits the "last visible" refusal. Each toggle
+# succeeds until only one is left standing, whose toggle is refused — which also parks the
+# selection on that row.
+private def leave_one_visible(o : TabsOverlay) : Nil
+  o.to_prefs.each_with_index do |(_, vis), i|
+    next unless vis
+    o.set_selected(i)
+    o.toggle_selected
+  end
+end
+
+describe "C5 · TabsOverlay on the Overlay seam" do
+  it "supplies the chrome the collapsed title/hint ladders read off it" do
+    OverlayHarness.new(TabsOverlay.new).assert_chrome(OverlayKind::Tabs, "TAB BAR")
+  end
+
+  it "applies the working copy on ↵ and discards it on esc" do
+    ov = TabsOverlay.new
+    ov.on_commit = -> { true } # mirrors the open-site's save_tabs
+    h = OverlayHarness.new(ov)
+    h.press(ENTER).should eq(:closed)
+    h.commits.should eq(1)
+
+    esc = OverlayHarness.new(TabsOverlay.new)
+    esc.press(ESC).should eq(:closed)
+    esc.commits.should eq(0) # esc must never persist the working copy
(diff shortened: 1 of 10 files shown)


hahwul:
Rebased onto `439ce2f` (C1 #373 + the #376 arity fix). Eight ladder conflicts in `runner.cr`, all the same shape — C1 and C5 each deleting their own arms from the seven shared `case @overlay` ladders — resolved as the union of both deletions. `MIGRATED_KINDS` in the dispatch spec took both batches' blocks, which is the merge the ledger was designed for.

Per the #376 postmortem, verified the **merge result** rather than trusting a green branch:

```
git fetch origin main && git rebase origin/main && crystal build --no-codegen src/main.cr   # clean
just test                                                                                   # 4106 examples, 0 failures
```

hahwul:
Rebased onto `bf282c8` (C3 #380 + #381). Conflicts were the ledger append plus two `runner.cr` ladder regions where C3 and C5 each deleted their own arms — resolved as the union of both deletions. One caveat for anyone resolving these mechanically: a blind intersection over the click-helper region left orphaned `end` fragments, because structural lines match across *different* methods on each side. Removed by hand and rebuilt.

**Verified against C3's now-migrated ConfirmDialog**, which was the cross-batch hazard I flagged on #355. `confirm()` captures the parent before `open_overlay`, and `restore_overlay(back, parent)` takes the `parent.key == kind` branch for both of my reset confirms, so my editors are restored as objects rather than as a phantom `@overlay` state. Live in the TUI:

- `Settings: Tabs` → `r` → CONFIRM → **esc lands back in TAB BAR**
- hide a tab (13/17 → 12/17) → `r` → `y` → back in TAB BAR at **13/17**, i.e. the action ran against the restored instance, not a dropped one
- PREFERENCES → ↵ Theme opener → SETTINGS → esc → **PREFERENCES** (nesting still intact post-rebase)

## One real finding from the adversarial review pass, fixed

`HostsOverlay#commit` and `EnvOverlay#commit` — the methods that parse the `IP host` / `KEY VALUE` add-row — **silently overrode `Overlay#commit : Bool`**, the base method the shell calls to run the injected closure and decide whether to close. Crystal has no `override` keyword, so nothing flagged it. Same class as the `Overlay#title` collision C2 hit in #370.

Inert today (neither editor ever returns a `:commit` outcome — both persist per mutation), but a landmine the moment one does: the shell would run the field parser instead of `on_commit` and read its truthy Symbol — `:empty` — as "close me". Renamed to `commit_entry`, with a comment at each saying why the obvious name is taken.

Added a regression example driving both through an `Overlay`-typed reference, which is how the shell holds them. Control-run: restoring the old name fails it.

`crystal build --no-codegen` clean against `bf282c8`; `just test` → **4195 examples, 0 failures**.

Holding the merge for C4 per the coordinator — I will rebase once more after it lands.

hahwul:
Second adversarial review pass done (both reviewers dogfooded the six modals in a live TUI, not just read the diff). Net: **no behaviour divergence in the migration**, one real pre-existing bug found and filed separately, two small things fixed here.

**Filed as #384, deliberately not fixed in this PR.** A confirm raised over another modal discards it without running its `on_close`, whenever `return_to:` does not name it — which the opt-in quit confirm always does (`return_to: :none`). Three visible symptoms: a stuck live theme preview, the nested pop-back into Preferences lost, and Preferences' unsaved-edits guard bypassed.

It is **not a regression**. On `22c35af`, `cancel_settings` was reachable only from `click_settings` and `handle_settings_key`'s esc/`^P`, and `settle_sub_editor` only from the key dispatch and the click ladder — the quit-confirm early-return in `handle_key` reached neither, so the preview stuck and the hop was lost there too. This PR preserves that exactly. What changed is visibility: C5 is the first batch to put real undo-work and a pop-back into `on_close`. Fixing it means deciding whether declining a quit should *restore* the displaced modal or drop it, and whether `close_or_warn` still gets to protect typed input — a product call, not something to slip into a behaviour-preserving refactor landing last into a file three batches share.

**Fixed here**, both mine:

- A comment on `confirm_section_reset` still claimed "raising a confirm only moves `@overlay` — `@active_overlay` still holds the card". That stopped being true when ConfirmDialog moved onto the seam in #380; the `parent` capture in `confirm` is now the only thing holding it. Rewritten, because that is precisely the sentence the next migration would trust.
- A stray blank line inside the `MODAL_OVERLAYS` literal, left by my rebase conflict resolution.

Everything else came back clean across both passes: no leaked or double-run closures, no reachable zombie state, the `return_to: :settings` / `:tabs` confirms restore the same object (verified live, accept *and* cancel), Preferences nesting keeps its focus and runs `refresh(:network)`, closure captures are all sound, and the hotkeys capture carve-out is byte-for-byte equivalent precedence to the original.

One reviewer also enumerated every `Input::Key` × char × modifier combination against each deleted Runner handler — 20,160 events per modal — and found **0 branch-selection divergences**, which covers the two reorderings I was least sure of (hoisting `^R` in the settings card, and collapsing the tabs `shift?` arms into a ternary).

`crystal build --no-codegen` clean against `bf282c8`; `just test` → 4195 examples, 0 failures. Still holding the merge for C4.

PR #387, opened 2026-07-26
docs(verb): correct the P0/P1 claims in the verb system and record the decision

### LINKED ISSUE #357: refactor(verb): decide ExecContext's fate — 266 abstract methods, one implementor
## Problem

`Verb::ExecContext` is an abstract class with **266 abstract methods** (42 in `verb/context.cr` + 224 across `verb/context/*.cr`) and **exactly one production implementor**: `Tui::Runner` (plus `spec/support/fake_context.cr`).

Its own doc comment claims the opposite:

```crystal
// src/gori/verb/context.cr:26-30
// The TUI App provides the concrete implementation; tests use a recording double.
// Keeping this interface thin is deliberate (P0): it is the entire surface
// verbs can affect, so it doubles as the action catalogue.
```

266 methods is not thin, and with one implementor there is no polymorphism to pay for the indirection.

## The promise that was not kept

`src/gori/verb.cr:5-8`:

> The verb system — gori's "same surface" core (P1). A single Definition is the one source of truth that drives a keybinding AND a command-palette entry (**and later an MCP tool + CLI subcommand**). No per-surface code paths.

The "later" never arrived. Verified:

```
$ grep -rn 'Verb::Registry\|Verb::Definition\|Verb::Scope' src/gori/cli src/gori/mcp
(no matches)
```

The `registry.` references in `src/gori/cli/` are `ProjectRegistry` (`run.cr:208`) and `Decoder.shared_registry` (`run/decoder.cr:131`) — unrelated. `Verbs.registry` is called from exactly one production site, `app.cr:54`, and handed to the TUI.

So `Verb::Scope`'s 30 enum values, `Category`, `mnemonic`, `section`, the keymap, and the OS profile all exist to serve one surface, while CLI (7.9k LOC) and MCP (8.2k LOC) declare their commands independently. #356 documents the resulting behavioural drift.

## This needs a decision, not a refactor

**Option 1 — finish P1.** Make `Verb::Definition` the source of truth for all three surfaces: CLI subcommands and MCP tools generated from the registry, with the argument schema that `verb.cr:99` notes as "intentionally absent this milestone — no verb takes arguments yet, and the field is additive later". Large, but it is the design as written and would delete a lot of parity-audit work permanently.

**Option 2 — scope it honestly to the TUI.** Accept that verbs are a TUI concern: rename to something surface-specific, drop the P1 claim from the comments, and collapse `ExecContext` into direct `Runner` calls (or a much smaller interface covering only what the palette and keymap actually need to introspect). Cheap, and it stops the abstraction from implying a guarantee it does not provide.

**Recommendation: Option 2**, unless CLI/MCP parity work is expected to continue at its current rate. The parity audits already shipped (see the `cli-mcp-parity` and `tui-cli-mcp-parity` work) show the cost of *not* having a shared definition — but they also show the gaps are mostly in argument surfaces and data serialization, which a verb registry with no argument schema does not solve. Option 1 is only worth it if the argument schema is built too.

## Scope of this issue

1. [ ] Decide between Option 1 and Option 2, and record the decision in `DESIGN.md` (from #353) under P1
2. [ ] Update or remove the P1 claim in `verb.cr:5-8` and the "thin interface (P0)" claim in `verb/context.cr:26-30` so the comments match reality either way
3. [ ] Execute the chosen option

Step 1 alone is worth landing first — the misleading comments are actively costing reviewer time, since they read as a description of the current state.

## Parallelism

- **Blocked by #355.** Both rewrite `runner.cr` structurally (`Runner` is the sole `ExecContext` implementor, and Option 2 removes the base class it inherits from). Better decided after #355 shrinks `Runner` from 5.5k to ~2.5k lines — the right answer may look different at that size.
- **Depends on #353** for a place to record the decision.
- Step 1 (the decision + comment fix) is independent and can land immediately, in parallel with everything.

hahwul:
Closes #357.

## What this is

Comments and docs only. No `.cr` code changed, no structure touched, no behaviour change.

Two comments described a system that does not exist:

- `src/gori/verb.cr:4-8` promised that one `Verb::Definition` drives "a keybinding AND a command-palette entry (and later an MCP tool + CLI subcommand)". The "later" never arrived and is not planned.
- `src/gori/verb/context.cr:26-30` called `ExecContext` "thin … deliberately (P0)" — for a 266-method interface.

Both read as descriptions of the present, which is what made them expensive: a reviewer takes them at face value.

## Measured on `main` at `57f1812` (re-verified in this branch, not copied from the issue)

| | |
|---|---|
| `ExecContext` abstract methods | **266** (42 in `verb/context.cr`, 224 in `verb/context/*.cr`) |
| lines of declarations | **478** |
| `Tui::Runner` (+ `runner/*.cr`) defines | **534**, implementing all 266, none missing |
| production implementors | **1** (`spec/support/fake_context.cr` is a recording double) |
| `Verb::Registry\|Definition\|Scope` in `src/gori/cli` + `src/gori/mcp` | **0 hits** |

One correction to the analysis comment on the issue: the declarations are **478** lines, not 986. 986 is `verb/context*.cr` (478) plus `spec/support/fake_context.cr` (1146)… which does not add up either; whatever produced it, 478 is what the tree has. Everything else in that comment reproduced exactly.

## The decision (Option 2, as settled on the issue)

**`ExecContext` is kept.** Collapsing it into direct `Runner` calls would delete the 266 declarations and the indirection, not the 266 implementations — the palette and keymap still have to invoke them. Large mechanical edit, zero behavioural gain, and it would destroy the two things the interface does buy: one enumerable catalogue of every action a verb can trigger, and a one-way dependency (`verb/` names no `Tui::` in code, per §2.1).

**CLI/MCP stay separate.** Not because the intents are TUI-shaped — 224 of the 266 are tool intents, surface-neutral in name — but because each is implicitly parameterised by TUI selection state. Verified at the call site, not inferred from the name: `repeater_send` reaches `repeater_controller.repeater_send`, whose first line is `return unless tab = current_repeater_tab`; `probe_rule_toggle` reaches `probe_controller.rules_toggle_selected`. A CLI or MCP caller has an id, not a selection. Making these callable elsewhere needs the argument schema `Verb::Definition` records as absent — that is the prerequisite, not a follow-up to registry wiring, and it is a project of its own across 224 intents.

## Changes

- `src/gori/verb.cr` — header now describes the TUI reality (keybinding + palette + space menu, all through `Definition#call`) and states plainly where it stops and why. The `Definition` comment now says the missing argument schema is load-bearing rather than "absent this milestone … additive later".
- `src/gori/verb/context.cr` — honest instead of flattering: wide, one implementor, no polymorphism being bought, and an explicit "do not read this as P0 minimalism", followed by what it actually buys.
- `src/gori/mcp.cr` — the only other comment in `src/` touching this question. It already pointed the right way; it now gives the precise reason (a verb names no target) and says the separation is settled, not pending.
- `DESIGN.md` §7 — new entry, `Refines: [P1](#p1). Issue #357.`, following the existing format, with the numbers and the rejected options.

Nothing else in `src/` or `docs/` asserts the three-surface promise (checked).

## Two notes for the reviewer

1. **Suggested retitle for #357:** "correct the P0/P1 claims in the verb system". It is no longer "decide ExecContext's fate" — the ExecContext-collapse option was measured and rejected for the reasons above.
2. **`DESIGN.md` §1 P1 is untouched, deliberately.** Its closing sentence still reads "that gap is real, known, and under decision in issue #357", which this PR settles. §7's own rule is to append rather than edit a principle's wording, so instead the new entry names that sentence and marks it settled. If you would rather P1 point straight at §7, that is a one-line follow-up — I did not want to edit a principle under a docs-only change.

## Verification

- `crystal build --no-codegen src/main.cr` — OK (this is also the proof that all 266 abstract methods are implemented; Crystal would reject the subclass otherwise)
- `just test` — **4265 examples, 0 failures, 0 errors**
- `crystal tool format --check` on the three touched `.cr` files — clean
- ameba on the touched files — 1 finding, pre-existing (`%w`-delimiters at `verb.cr:64`, untouched by this diff; it only moved line numbers)
- No specs added: there is no behaviour here to assert, and no spec reads `DESIGN.md`

`just check` does fail tree-wide, on `spec/proxy/tls/tunnel_spec.cr` and `src/gori/proxy/upstream.cr` — pre-existing Crystal 1.21.0 format drift in files this branch does not touch.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -416,6 +416,52 @@ A surface still owns its own error wording: `Fuzz::PlanError` carries a machine-
 `reason`, and each surface renders the sentence naming its own flags (`--auto` / `auto:true` /
 `^A params`). Sharing the assembly must not flatten three different vocabularies into one.
 
+### 2026-07-26: the verb registry is a TUI concern, and `ExecContext` is a catalogue
+
+Refines: [P1](#p1). Issue #357.
+
+Two comments described a system that does not exist. `src/gori/verb.cr` promised that one
+`Verb::Definition` drives "a keybinding AND a command-palette entry (and later an MCP tool +
+CLI subcommand)", and `src/gori/verb/context.cr` called `ExecContext` "thin … deliberately
+(P0)". Neither the later nor the thin was true: `grep -rn 'Verb::Registry\|Verb::Definition'
+src/gori/cli src/gori/mcp` returns nothing, and the interface declares 266 abstract methods.
+Both read as descriptions of the present, which cost reviewers time. The comments are now
+corrected; the structure is deliberately unchanged.
+
+Measured on `main` at `57f1812`:
+
+- `ExecContext` requires **266** abstract methods: 42 in `verb/context.cr` and 224 more spread
+  across the twenty per-tool files in `verb/context/`, which exist only to hold declarations.
+- `Tui::Runner` (with its `runner/*.cr` mixins) defines **534** methods and implements all
+  266, none missing. It is the only production implementor; `spec/support/fake_context.cr` is
+  a recording double.
+- Only the 42 root ones are app chrome and cross-tool actions. The other 224 are tool intents
+  grouped by tool (repeater 28, issues 23, probe 22, history 17, fuzzer 17, jwt 13 …) —
+  surface-neutral in name.
+
+Collapsing `ExecContext` into direct `Runner` calls was considered and rejected. It would
+delete the 266 declarations and the indirection, not the 266 implementations, which the
+palette and keymap still have to invoke — a large mechanical edit for no behavioural gain. It
+would also destroy the two things the interface does buy: one enumerable catalogue of every
+action a verb can trigger, and a one-way dependency, since `verb/` names no `Tui::` in code
+([§2.1](#s2-1)). Keeping a 266-method abstraction is not a [P0](#p0) minimalism claim, and it
+should stop being written up as one.
+
+Wiring CLI and MCP into the registry was also rejected, for a reason worth recording because
+it is not obvious from the method names. Those 224 tool intents are surface-neutral in name
+and TUI-coupled in semantics: `repeater_send` means "send the ACTIVE sub-tab",
+`probe_rule_toggle` means "toggle the HIGHLIGHTED row". A CLI or MCP caller has no selection,
+only an id. Making them callable from another surface therefore requires an argument schema
+so an intent can name its target — the field `Verb::Definition` records as absent. That
+schema is the prerequisite, not a follow-up to the wiring, and it is a project of its own
+across ~224 intents.
+
+So P1's reach stays where [P1](#p1) already describes it: one execution path inside the TUI,
+parity elsewhere by calling the same engines. Read P1's closing sentence — "that gap is real,
+known, and under decision in issue #357" — as settled by this entry: the gap is deliberate, and
+what would close it is the argument schema, not registry wiring. If CLI/MCP parity work resumes
+at the rate it ran before, revisit, but open the argument-schema issue first.
+
 ---
 
 *Keep this document honest against the code. When you change a subsystem it describes, update
--- a/src/gori/mcp.cr
+++ b/src/gori/mcp.cr
@@ -8,10 +8,12 @@ require "./mcp/install"
 module Gori
   # The `gori mcp` server: exposes captured data + the repeater engines to an AI
   # client over the Model Context Protocol (JSON-RPC 2.0 on stdio). It talks
-  # straight to Store + Repeater rather than the TUI verb registry, because verbs
-  # drive a live ExecContext (UI state) and have nothing to offer a headless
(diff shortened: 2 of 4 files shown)


hahwul:
Rebased onto `00df56a` (main had moved by #385 and #386) and re-verified the merge result: `crystal build --no-codegen src/main.cr` OK, `just test` 4265 examples / 0 failures, and the counts still read 266 abstract / 534 defined.

One self-correction while re-measuring, worth naming because it is the exact failure this PR is about. The §7 entry originally recorded "478 lines of declarations" — a number this PR's own comment additions push to 489, so it would have been wrong on the first main that carried it. Replaced with a structural statement (42 in `verb/context.cr`, 224 across the twenty per-tool files in `verb/context/`) that cannot rot without a real structural change. The method counts stay: those go stale only when someone genuinely adds an intent, which is when the entry *should* be re-read.

Issue #382, opened 2026-07-25
CI never builds the merge result, so semantic conflicts between green PRs break main


hahwul:
## What happened

`origin/main` did not compile for a stretch today. Two PRs that were each CI-green on their own branch combined into a broken tree:

- **#372** (`22c35af`) split the shell's overlay teardown into `close_active_overlay(ov : Overlay)`, which drops the modal **and** runs its `on_close` pop-back, and `leave_overlay`, which drops it **without**.
- **#371** (`b377a0f`) added a call site written against the old zero-argument signature:

```crystal
// src/gori/tui/runner.cr:3864
ov.on_palette = -> { close_active_overlay; open_palette }
```

```
Error: wrong number of arguments for 'Gori::Tui::Runner#close_active_overlay' (given 0, expected 1)
Overloads are:
 - Gori::Tui::Runner#close_active_overlay(ov : Overlay)
```

Both PRs touched `runner.cr` in different places, so git merged them cleanly with no conflict to resolve. Fixed in #376.

Two more PRs (#375, #377) then failed CI for a reason that had nothing to do with their own diffs, and the authors had to be told not to "fix" it locally — which would have produced a second, conflicting edit to the same line.

## Why CI could not catch it

`.github/workflows/ci.yml` triggers on `pull_request` and on `push` to `main`. So it builds:

- each PR branch, in isolation
- `main` **after** the merge has already landed

There is no build of the *prospective merge result* before it becomes `main`. A semantic conflict — one PR changes a signature, another adds a call site, in different hunks — is invisible to both.

This is not specific to the overlay work. It is a property of any repository merging concurrent PRs that touch a shared file, and it gets sharply more likely with parallel authors on one subsystem. `src/gori/tui/runner.cr` had five PRs in flight against it at once.

## Options

**1. GitHub merge queue (`merge_group` trigger).** Add `merge_group:` to `ci.yml`'s `on:` and enable the queue on `main`. GitHub then builds each PR *as merged with the current tip* and refuses to land it if that build fails. This is the direct fix for exactly this failure mode. Costs one extra CI run per merge, and requires the repo to allow a queue.

**2. Require branches to be up to date before merging.** A branch-protection setting. Cheaper, no workflow change, and it would have caught this case — but it serializes merges by hand (every PR rebases after every other PR lands), which is painful with five concurrent authors and is exactly the churn the queue automates.

[...]

PR #388, opened 2026-07-26
fix(tui): restore a modal displaced by a :none confirm instead of dropping it (#384)

### LINKED ISSUE #384: A confirm raised over another modal discards it without running its close hook
## Summary

A confirm raised over another modal **discards that modal without running its close
hook**, whenever `return_to:` does not name it. The reachable trigger is the opt-in quit
confirm (settings → General → *Confirm before quit* = on), which uses the default
`return_to: :none` and so hits every modal in the app.

`src/gori/tui/runner.cr`, `restore_overlay`:

```crystal
private def restore_overlay(kind : OverlayKind, parent : Overlay?) : Nil
  return open_overlay(parent) if parent && parent.key == kind
  restorable = kind.none? || kind.detail? || MODAL_OVERLAYS.includes?(kind)
  @overlay = restorable ? kind : OverlayKind::None
end
```

By the time this runs, `close_active_overlay` has already called `leave_overlay`, so
`@active_overlay` is nil. When the first branch does not match, `parent` is dropped and
nothing will ever call `parent.on_close`. That contradicts the contract in `overlay.cr`:
*"What the shell runs AFTER this overlay closes, whether it committed or cancelled."*

## Reproduce

Turn on settings → General → **Confirm before quit** first. Then, with the theme card up:

1. `^P` → `Settings: Theme`
2. `↓ ↓` to preview a different theme (the whole UI recolours — that is the live preview)
3. `^C`, then `n` to decline the quit

The card vanishes and the previewed theme **stays applied**, while `settings.json` still
holds the old one. Nothing on screen offers a way back; you have to re-open the theme card
and press esc to trigger the revert.

Two more symptoms, same root cause:

- **Nested pop-back lost.** `^P` → `Settings` → *Network & Tabs* → `Tabs` row `↵` → `^C` →
  `n` lands on the bare tab body instead of back in Preferences. Same for
  hosts / env / hotkeys / theme.
- **Preferences' unsaved-edits guard bypassed.** Open `Settings: Network`, edit Bind Port
  (the section header shows `● unsaved`), `^C`, `n` — the modal is dropped with no warning
  and the edit is gone. Esc and click-away both go through `PreferencesView#close_or_warn`;
  this path never reaches it.

## Not a regression

This predates the Overlay migration. On `22c35af`, `cancel_settings` was reachable only
from `click_settings` and `handle_settings_key`'s esc/`^P`, and `settle_sub_editor` only
from the key dispatch and the click ladder — the quit-confirm early-return in `handle_key`
reached neither. So the theme preview stuck and the Preferences hop was lost there too.

What changed is visibility: #379 is the first batch to put real undo-work (the theme
revert) and the nested pop-back into `on_close`, so the hole now has consequences a user
can see. #379 preserves the old behaviour exactly, which is why it is not fixing this.

## Why this is not folded into #379

The fix is a behaviour change, and the right behaviour is a product call I should not make
unilaterally in a refactor that is landing last into a file three batches share:

- Should declining the quit **restore** the displaced modal, rather than run its `on_close`
  and drop it? That is arguably what the user expects from "no, do not quit" — the whole
  stack comes back, not just the theme un-previewing.
- If instead it drops, should `close_or_warn` still get a chance to protect unsaved edits?
  A quit confirm silently discarding typed input is the worst of the three symptoms.

My suggestion is **restore, not drop**: capture `parent` (as `confirm` already does) and
re-open it whenever `return_to:` is `:none` *and* a parent was displaced — treating
`:none` as "I did not ask to change where you were", which is what every one of these
call-sites means. `restore_overlay`'s existing first branch already does exactly this for
a named parent. Falling back to running `parent.on_close` would also close the hole, but
gives up the stack and still throws away the unsaved edits.

Either way it wants its own PR with the three repro paths as specs.

hahwul:
## Summary

A confirm raised **over** another modal with the default `return_to: :none` discarded that modal without running its `on_close`. `restore_overlay` only re-opened the parent when `return_to:` named its kind (`parent.key == kind`); a `:none` confirm fell through the first branch and dropped the parent — teardown and all.

The one reachable trigger is the opt-in **quit confirm** (settings → General → *Confirm before quit*), which fires from `Runner#handle_key` while **any** modal is up. Declining it produced the three symptoms in #384:

- a live theme preview stuck applied (its revert `on_close` never ran),
- a nested sub-editor gone, its pop-back to Preferences lost,
- an unsaved Preferences edit silently discarded, bypassing `close_or_warn`.

## The design decision: restore, not drop

The issue laid out two options for a modal a confirm displaces: (a) preserve and restore it, or (b) close it properly by running its `on_close` at displacement time. **This PR picks (a).**

A displaced modal was *never asked to close*. The confirm interrupted the user (a global ^C/^D), it did not cancel the modal. Running the modal's `on_close` would treat that interruption as a cancel and fire teardown that is wrong for "no, don't quit": it would revert a live theme preview the user was still working on, and — worst of the three — discard a sub-editor's unsaved edits with no warning. Restoring brings the modal back exactly as it was, edits and all, which is what "no, don't quit" means. This also matches the issue author's own recommendation.

Restore is done by re-opening the **captured parent object**, which sets both `@active_overlay` and `@overlay`, so it cannot produce the phantom-overlay state the old comment warns about (a migrated kind restored by `@overlay` alone renders nothing and captures no keys). The state-restore fallback keeps its phantom guard for the parent-less case.

## The change

```crystal
-      return open_overlay(parent) if parent && parent.key == kind
+      return open_overlay(parent) if parent && (parent.key == kind || kind.none?)
```

`:none` now means "I did not ask to change where you were": over a modal it restores that modal; with nothing up it still lands on the bare body. The named-parent path (RESET SETTINGS / RESET TAB BAR raising a confirm with `return_to: :settings` / `:tabs` from inside the card) is unchanged. I audited every `confirm(` call-site: the only `:none`-over-a-migrated-modal path is the global quit confirm (checked *before* modal dispatch, so no controller confirm ever fires over a modal); every other `:none` confirm either has no modal up or drops its own modal first (e.g. `submit_ca_import`). The `restore_overlay` and `confirm` comments were rewritten to describe this accurately.

## Tests

`spec/tui/overlay_nesting_spec.cr`'s `NestShell` double gains faithful mirrors of `Runner#confirm` + `#restore_overlay` (the Runner needs a live tty and can't be unit-tested, so the file already mirrors its overlay-state machine; this keeps that discipline), driving the **real** `ConfirmDialog` so `y`/`n`/esc route exactly as the shell does. New examples:

- **quit-confirm decline** (`:none` over a migrated modal): restores the modal, and its `on_close` never runs — the #384 fix.
- **quit-confirm accept**: restores the modal *first*, then runs the quit action (the restore-then-act order History relies on).
- **named-parent** (`return_to: :settings`): still restores the object — regression guard for RESET SETTINGS/TAB BAR.
- **bare body**: a `:none` confirm that displaced nothing lands on `None`.
- **phantom guard**: a migrated kind with no object restores to `None`, not a phantom; an unmigrated `MODAL_OVERLAYS` member still restores by state.

**Control-run:** reverting `|| kind.none?` in the mirror turns the two `:none`-over-modal examples red while the three regression guards stay green — the intended split. Full suite: `crystal spec` → 4270 examples, 0 failures. Rebased onto current `main` and re-verified `crystal build --no-codegen src/main.cr` (the merge result) builds.

Closes #384

### DIFF
--- a/spec/tui/overlay_nesting_spec.cr
+++ b/spec/tui/overlay_nesting_spec.cr
@@ -70,6 +70,10 @@ private class NestShell
   property overlay = OverlayKind::None
   getter active : Overlay?
 
+  # Mirrors Runner::MODAL_OVERLAYS — the states restore_overlay is allowed to name without
+  # an object behind them. Keep in step (it is deliberately tiny; migrated modals are absent).
+  MODAL_OVERLAYS = {OverlayKind::Palette, OverlayKind::TabsMore}
+
   def open_overlay(ov : Overlay) : Nil
     @active = ov
     @overlay = ov.key
@@ -91,6 +95,32 @@ private class NestShell
   def active_overlay : Overlay?
     ov = @active
     ov if ov && @overlay == ov.key
+  end
+
+  # Mirror of Runner#restore_overlay (runner.cr). Keep in step — a spec that passes against
+  # a drifted copy proves nothing about the real shell. The `|| kind.none?` clause is the
+  # #384 fix: a :none confirm displacing a modal restores it, rather than dropping it.
+  def restore_overlay(kind : OverlayKind, parent : Overlay?) : Nil
+    return open_overlay(parent) if parent && (parent.key == kind || kind.none?)
+    restorable = kind.none? || kind.detail? || MODAL_OVERLAYS.includes?(kind)
+    @overlay = restorable ? kind : OverlayKind::None
+  end
+
+  # Mirror of Runner#confirm's overlay-state wiring (runner.cr): capture the displaced
+  # parent BEFORE open_overlay overwrites it, then wire the dialog to restore-first-act-
+  # second on close. Uses the real ConfirmDialog so `y`/`n`/esc route exactly as they do
+  # in the shell. `return_to` mirrors the production Symbol seam.
+  def confirm(*, return_to : Symbol = :none, &action : -> Nil) : Nil
+    ov = ConfirmDialog.new("CONFIRM", "quit?")
+    back = OverlayKind.from_sym(return_to)
+    parent = active_overlay
+    accepted = false
+    ov.on_commit = -> { accepted = true; true }
+    ov.on_close = -> {
+      restore_overlay(back, parent)
+      action.call if accepted
+    }
+    open_overlay(ov)
   end
 
   def press(k : Termisu::Input::Key) : Nil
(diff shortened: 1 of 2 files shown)


hahwul:
Coordinator: verified, and the reasoning is better than the framing I gave you.

I laid out two options and deliberately did not pick one. The argument you found is the one that decides it and I had not made it: **a displaced modal was never asked to close.** The confirm interrupts the user; it does not cancel the modal. Running `on_close` at displacement would treat an interruption as a cancel and fire teardown that is wrong for "no, don't quit".

You also found a third symptom the issue did not report, and it is the worst of the three: an unsaved Preferences edit **silently discarded, bypassing `close_or_warn`**. That is data loss, and on its own it rules out option (b).

Two things I want to note because they were not in your brief and you did them anyway:

- **You rebased onto current `main` and re-ran `crystal build --no-codegen` on the merge result.** That is exactly the failure class that broke `main` earlier today (#382) — two PRs green in isolation, different hunks, clean merge, no compile. Checking it unprompted is the right instinct.
- **Restoring by re-opening the captured parent object** sets both `@active_overlay` and `@overlay`, so the phantom-overlay state the old comment warns about is structurally unreachable rather than merely avoided. The state-restore fallback keeping its guard for the parent-less case is the right split.

Control-run is documented and the split is the intended one — reverting `|| kind.none?` reddens the two `:none`-over-modal examples while the three regression guards stay green.

Merge when CI is green.

PR #389, opened 2026-07-26
fix(discover): apply the Layer-2 scope gate to the seed and its two derived well-known paths (#364)

### LINKED ISSUE #364: Discover's seed + robots.txt/sitemap.xml bypass the Layer-2 scope gate (3 requests)
Surfaced by the code-review pass on #354. `src/gori/discover/**` was outside that issue's file boundary (`git diff origin/main -- src/gori/discover/` is empty on that branch), so it was flagged rather than fixed.

## The gap

`Discover::Engine#seed_frontier` pushes the seed URL plus two derived well-known paths straight onto the frontier with no `bounded_url` call:

```crystal
// src/gori/discover/engine.cr:322-330
private def seed_frontier : Nil
  @seen << Url.visit_key(@seed_parts)
  if @config.spider?
    @crawl_enqueued += 1
    @frontier << Task.new(TaskKind::Crawl, Url.normalize(@seed_parts), 0, Source::Seed)
    root = Url.origin(@seed_parts)
    @frontier << Task.new(TaskKind::Fetch, "#{root}/robots.txt", 0, Source::Robots)
    @frontier << Task.new(TaskKind::Fetch, "#{root}/sitemap.xml", 0, Source::Sitemap)
  end
```

Every other enqueue path does gate — `bounded_url` is applied at `engine.cr:519` and `:541`. The file's own comment at `:349` already acknowledges a related bypass ("Bypasses bounded_url/scope the same way …"), so this is known-adjacent territory rather than a surprise.

Blast radius is **3 requests**: seed, `robots.txt`, `sitemap.xml`. Everything discovered afterwards is gated normally.

## Why it is not obviously a bug

The seed is what the operator typed, so arguably it belongs in the same category as the TUI's `Outbound.interactive` (no up-front gate because a human chose the target — see `DESIGN.md` §7).

`robots.txt` and `sitemap.xml` are weaker: they are **derived**, anchored on `Url.origin`, and on a path-scoped run confined to a subtree they sit outside that subtree by construction. The code deliberately wants them there for calibration (`engine.cr:340-343` explains why), but path-confinement and the scope/Sandbox gate are separate concerns, and only the former was reasoned about.

## What to decide

- Should the seed itself be gated, or treated as `interactive` like the TUI?
- Should `robots.txt` / `sitemap.xml` pass through `Outbound#sweep_block` (Sandbox + EXCLUDE) even when the seed is trusted? Sandbox mode's documented promise is "blocks ALL out-of-scope traffic", which currently does not hold for these three requests.

Whatever is chosen, record it in `DESIGN.md` §7 alongside the #354 decision so the asymmetry is deliberate rather than inherited.

hahwul:
Closes #364.

## The decision

The issue asked for a product decision before a patch, so here is the reasoning first.

**The seed is "interactive" for Layer 1, and stays ungated there.** That is not a new judgement
call — it is the one the codebase already made. `Outbound.interactive` exists precisely because
"the operator typed this target" answers the include question, and each discover surface has
already applied its own Layer-1 verdict before the engine runs (`Outbound.agent` refuses an
out-of-scope seed, `Outbound.cli` refuses one on a configured project, `.interactive` waives by
name). Re-asking the include question inside the engine would only re-ask what the surface just
answered.

**But Layer 2 was never part of that bargain, and all four requests now pass it.** Sandbox's
promise in DESIGN.md §3 — "a request to a host that is not allowlisted is blocked outright" —
carries no "unless the operator typed it" clause. `Outbound.interactive`'s own doc comment says
so explicitly: *"the operator typed the target, so there is no up-front gate — but Sandbox still
hard-stops every send."* That promise did not hold for these requests. "The operator chose this
target" was never an argument about Sandbox.

Explicit EXCLUDE rules come along with Sandbox (`sweep_block` semantics, not `send_block`),
because discover is the most automated sweep gori has and every *other* URL in the run is already
judged by that same predicate.

**robots.txt / sitemap.xml keep waiving Layer 1, containment and the path confine.** They are
derived from a seed the operator was already authorised to hit, they live at that origin by
construction, and the calibration reason for reaching outside a path-confined subtree
(`engine.cr` ~340) is sound and unchanged. Path-confinement and the scope gate really are
separate concerns — the fix keeps the former waived and closes the latter.

### Exposure was not uniform, and that is what settles it

A Layer-1 `in_scope` verdict already implies a clean Layer 2 (`Scope#sandbox_blocks?` and
`Outbound#evaluate` both route through `allowlisted_unlocked?`). So for the **seed** the gap only
ever opened where Layer 1 was waived: the TUI and any `--allow-unscoped` run. The three
**derived** requests were exposed on every surface, because they are anchored on `Url.origin` and
a path-scoped include rule never covers them.

### A fourth bypass, larger than the three in the issue

The issue scoped the blast radius at 3 requests. It is 3 + `calibrate_probes + extensions.size`:
`enqueue_seed_only_calibration` fires the origin-root soft-404 calibration, whose bogus probes are
real requests and which bypassed the gate the same way. That is now gated too, on its directory.

## What changed

- `Engine#initialize` refuses the run outright when Layer 2 blocks the seed — a terminal
  `ErrorEvent` (`Engine::SEED_BLOCKED`), no `DoneEvent`. Loud, not a skipped enqueue: a blocked
  seed blocks everything derived from it, so the alternative is a run that ends with zero findings
  and no reason, which reads as "there is nothing there" rather than "gori sent nothing" (P4).
- `enqueue_well_known` gates robots.txt / sitemap.xml. A blocked one is skipped silently — unlike
  the seed, the crawl is still meaningful without it.
- `enqueue_seed_only_calibration` gates the origin-root calibration on its directory.
- DESIGN.md §7 records the decision alongside the #354 entry.

`@seed_calibration_dir` is deliberately still set even when that calibration is refused. Dropping
it was my first draft and it was wrong: without it a robots/sitemap outcome falls back to
`record_page`, whose raw-status trust reports both as findings on a 200-everything server — the
exact false positive the calibration exists to prevent. `spec/discover/engine_spec.cr` pins that.

## Verification

- `just test` — 4273 examples, 0 failures, on the rebased tree.
- `crystal build --no-codegen src/main.cr` after the rebase (CI does not build the merge result).
- `crystal tool format --check` and `ameba` clean on the three touched files (0 new findings;
  the file's 8 pre-existing `%w()` style findings are untouched).
- Every new spec was **control-run against the unfixed engine** and fails there: 3 in
  `engine_spec.cr`, 2 in `plan_spec.cr`, plus the false-positive regression spec run against my
  own bad draft. The `plan_spec` pair drives a real `Gori::Scope` with Sandbox enabled through a
  real TCP listener and asserts on the request heads the origin actually received.

## Deliberately not closed here

Found while reviewing this diff, all **pre-existing on main** and each a separate decision. Filed
rather than folded in, to keep a security-gate change reviewable:

- Brute-force and calibration probes are authorised by their *directory*, not per URL, so a
  `string`/`regex` EXCLUDE matching a child but not its parent does not stop them (~278 requests
  per directory on one waiver).
- `Plan.resolve_policy` hands an unconfigured scope an `OpenScope`, so Layer 2 is absent entirely
  on a project with Sandbox on and no rules.
- A crawled `<a href>` carrying a raw CRLF splices a second, fully attacker-chosen request into
  the request line. `Plan.build` already guards exactly this for the seed via
  `Headers.safe_value?`; the guard was never applied to crawled links.
- A single-segment seed with no trailing slash (`http://t/api`) sets `@seed_calibration_dir`
  while the path confine refuses the matching Calibrate task, so robots/sitemap are fetched and
  their outcomes are never recorded.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -462,6 +462,65 @@ known, and under decision in issue #357" — as settled by this entry: the gap i
 what would close it is the argument schema, not registry wiring. If CLI/MCP parity work resumes
 at the rate it ran before, revisit, but open the argument-schema issue first.
 
+### 2026-07-26: Discover's seed waives Layer 1, never Layer 2
+
+Refines: [P4](#p4). Issue #364, surfaced by the review of #354.
+
+`Discover::Engine#seed_frontier` put the seed, `<origin>/robots.txt`, `<origin>/sitemap.xml` and
+the origin-root soft-404 calibration straight onto the frontier with no `bounded_url` call.
+Every URL the crawl derived afterwards was gated normally, which is why it read as a deliberate
+seed exemption rather than a hole, but only the *path-confinement* half had ever been reasoned
+about (well-known paths live at the origin, so a run confined to `/app/` has to step outside its
+subtree to find them). The scope gate rode along with it by accident.
+
+The exposure was not uniform across the four, and the difference is what settles the decision. A
+Layer-1 `in_scope` verdict already implies a clean Layer 2, because `Scope#sandbox_blocks?` and
+`Outbound#evaluate` both route through `allowlisted_unlocked?` — so for the **seed** the gap only
+ever opened where Layer 1 was waived, which is the TUI and any `--allow-unscoped` run. The three
+**derived** requests were exposed on every surface, since they are anchored on `Url.origin` and a
+path-scoped include rule never covers them.
+
+The two-layer split in [§3](#s3) already answers this, and the answer is asymmetric on purpose:
+
+- **Layer 1 stays waived for all four.** The seed is what a human typed, which is the same
+  argument `Outbound.interactive` makes for the TUI, and every surface has already made that
+  decision before the engine runs (`Outbound.agent` refuses an out-of-scope seed, `.cli` refuses
+  one on a configured project, `.interactive` waives by name; the CLI and MCP enforce the verdict
+  right after `Plan.build`, before a single send). `robots.txt` and `sitemap.xml` inherit it:
+  they are derived from a seed the operator was already authorised to hit and live at that same
+  origin by construction. Re-asking the include question here would only re-ask what the surface
+  just answered, and on a path-scoped include rule it would break the calibration on every run
+  that has one.
+- **Layer 2 now applies to all four.** Sandbox's documented promise in [§3](#s3), that a request
+  to a host which is not allowlisted is blocked outright, carries no "unless the operator typed
+  it" clause, and `Outbound.interactive`'s own contract already says Layer 2 still hard-stops the
+  send. "The operator chose this target" was never an argument about Sandbox. Explicit EXCLUDE
+  rules come with it (`sweep_block` semantics, not `send_block`): discover is the most automated
+  sweep gori has, and its every other URL is judged by the same predicate.
+
+Three specifics worth recording:
+
+- **A blocked seed fails the run, loudly.** The verdict is taken in `Engine#initialize` and
+  `Engine#start` emits it as the run's sole terminal `ErrorEvent` (`Engine::SEED_BLOCKED`); it is
+  not a skipped enqueue. A blocked seed blocks everything derived from it, so the alternative is
+  a run that finishes with zero findings and no reason, which an operator reads as "there is
+  nothing there" rather than "gori sent nothing" ([P4](#p4)). A blocked `robots.txt`/`sitemap.xml`
+  is skipped silently instead: the crawl is still meaningful without it.
+- **A gated calibration still routes its two dependants.** `@seed_calibration_dir` is set whether
+  or not the Calibrate task survives the gate. Without it a robots/sitemap outcome falls back to
+  `record_page`, whose raw-status trust reports both as findings on a 200-everything server,
+  which is the exact false positive the calibration exists to prevent. With it and no baseline
+  they go uncounted: no baseline, no claim.
+- **The gate is the engine's injected `ScopePolicy`, not an `Outbound`.** `StoreScope#allowed?`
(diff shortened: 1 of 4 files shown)


hahwul:
Coordinator: verified, and this corrected my framing of the issue in two places.

I asked "is the seed interactive?" as one question. Splitting it by layer is the right move and settles it: the seed **is** interactive for Layer 1 — that is the judgement the codebase already made, and re-asking inside the engine would re-ask what each surface just answered — while **Layer 2 was never part of that bargain**. `Outbound.interactive`'s own doc comment says so outright ("the operator typed the target, so there is no up-front gate — but Sandbox still hard-stops every send"), and `DESIGN.md` §3 carries no operator-typed exemption. "The operator chose this target" was never an argument about Sandbox. I had those two conflated in the issue text.

Two things I got wrong that you fixed:

- **Exposure was not uniform.** Since a Layer-1 `in_scope` verdict already implies a clean Layer 2 via `allowlisted_unlocked?`, the seed's gap only ever opened where Layer 1 was waived (TUI, `--allow-unscoped`), while the three derived requests were exposed on **every** surface because `Url.origin` anchoring means a path-scoped include never covers them. My issue treated all four as one bypass.
- **The blast radius is bigger than the "3 requests" I wrote.** `enqueue_seed_only_calibration` fires origin-root soft-404 calibration whose bogus probes are real requests through the same hole, so it is 3 + `calibrate_probes + extensions.size`. Gating that too is correct.

Keeping path-confinement waived while closing the scope gate is exactly the separation the issue asked for and the calibration rationale at `engine.cr` ~340 survives intact.

This was the one issue in the batch where I deliberately withheld the answer. Withholding it produced a better one than I would have written. Merge when you are ready.

Issue #439, opened 2026-07-26
feat(cli,settings): --config PATH and settings export/import for shareable profiles


hahwul:
### Background

The settings file has exactly one location: `Settings.path` → `File.join(Paths.home_dir, "settings.json")`, where `home_dir` is `$GORI_HOME` or `~/.gori` (`src/gori/paths.cr:9`). `gori settings [--edit]` (`src/gori/cli.cr:132`) prints or opens that path. There is no `--config` flag, and no export/import (`rg -i 'export.*settings|import.*settings' src/` → 0 hits).

Burp exposes config as a first-class portable artifact: JSON export/import, a config library, and `--config-file` / `--user-config-file` on the command line.

### Problem

**Switching profiles means moving directories.** The only way to run gori with a different configuration is to relocate the entire `GORI_HOME` — which also relocates the CA, the project databases, themes, and wordlists. So "the same projects with a different network profile" is not expressible, and neither is "my normal setup, but with the corporate upstream proxy" as a one-line switch.

**CI and headless runs cannot be configured declaratively.** This is the sharpest version of the gap, because headless operation is a deliberate strength here — `gori run`, `gori mcp`, and the MCP seam are all built out. But a headless run can only be configured through CLI flags and whatever happens to be in `~/.gori/settings.json`. A reproducible pipeline wants a checked-in config file passed explicitly. Right now the workaround is to construct a throwaway `GORI_HOME` and copy the CA into it.

**Nothing is shareable.** A team cannot distribute scan rules, hostname overrides, hotkeys, or a theme as a reviewable artifact. They are entries inside a single machine-local file that also holds decoder session state and a plaintext env var list.

### Proposal

Two independent pieces; the first is small and worth doing on its own.

**1. `--config PATH`.** Point `Settings.path` at an explicit file for one process, orthogonal to `GORI_HOME`. Precedence: `--config` > `$GORI_CONFIG` > `$GORI_HOME/settings.json`. Available on the TUI launch path, `gori run`, and `gori mcp`.

**2. Export / import of a settings subset.**

```
gori settings export [--sections network,scan_rules,hotkeys] [-o profile.json]
gori settings import profile.json [--sections …] [--dry-run]
```

Wiring notes:

[...]

Issue #437, opened 2026-07-26
feat(proxy): configurable upstream TLS protocol versions and cipher suites


hahwul:
### Background

Upstream TLS has exactly one knob: `Settings.verify_upstream` (`src/gori/settings/network.cr`), a boolean. `Upstream.client_context` (`src/gori/proxy/upstream.cr:359`) builds an `OpenSSL::SSL::Context::Client` with stdlib defaults, sets `verify_mode = NONE` when verification is off, applies the system trust store (`apply_system_trust`, `:311`), and sets ALPN. No protocol floor/ceiling, no cipher list, no signature-algorithm or curve control.

Burp exposes the negotiated protocols and cipher suites directly ("use custom protocols and ciphers"), plus options for permissive handshakes with broken servers.

### Problem

**Legacy targets are unreachable.** A device or appliance that only offers TLS 1.0/1.1, or only a cipher suite the distribution's OpenSSL policy disables at `SECLEVEL=2`, fails the handshake with no way to relax it. `verify_upstream: false` does **not** help — it turns off certificate *verification*, not protocol/cipher negotiation. So the operator's only workaround is to rebuild against a different OpenSSL configuration, which is not a workaround.

**Handshake-fingerprint work is impossible.** Being able to pin the offered protocol/cipher list is how one investigates TLS-fingerprint-based blocking (JA3-style) or reproduces a client-specific handshake. Not being able to change the ClientHello at all forecloses that class of testing.

Note this is separate from the `verify_upstream` toggle already documented as re-syncing the running proxy, prober, and Repeater/Fuzzer/Miner senders live — the same live-resync behaviour is the model these new knobs should follow.

### Proposal

Extend the `network` section:

```json
{
  "network": {
    "tls_min_version": "tls1.2",
    "tls_max_version": "",
    "tls_ciphers": "",
    "tls_permissive": false
  }
}
```

- `tls_min_version` / `tls_max_version`: `"tls1.0"`…`"tls1.3"`, empty = stdlib default. Sets `OpenSSL::SSL::Context::Client#add_options` / the version bounds.
- `tls_ciphers`: an OpenSSL cipher string, empty = default. Setting it should also lower the security level when the string requires it, or the setting will appear to be ignored — `@SECLEVEL=0` behaviour needs to be explicit and documented rather than surprising.
[...]

Issue #434, opened 2026-07-26
feat(proxy): upstream connection rules — per-host routing, SOCKS5, proxy auth, and client TLS certificates


hahwul:
### Background

The entire outbound path is configured by **one string**: `Settings.upstream_proxy` (`src/gori/settings/network.cr`), plus a single per-project override (`net.upstream_proxy`). `Settings.upstream_proxy_addr` parses it into `{host, port}`, and every active sender reaches the wire through `Upstream.dial`, which CONNECT-tunnels through it when set.

That single scalar is the whole configuration surface for upstream connectivity. Burp covers the same ground with an upstream-proxy **rules table** (destination-host patterns, per-rule credentials, separate SOCKS settings) plus platform authentication and per-host client certificates; Caido has per-host upstream configuration too.

### Problem

Four distinct gaps, all rooted in the scalar:

**1. No per-destination routing.** "Route `*.corp.internal` through the internal proxy, everything else direct" cannot be expressed. There is also no `no_proxy`-style exception list (`rg -i 'no_proxy|proxy_bypass' src/` → 0 hits). The only granularity is global vs. per-project, and a project is usually *one* engagement spanning both internal and external hosts.

**2. No upstream proxy authentication.** `src/gori/proxy/upstream.cr:195` writes the tunnel request as:

```crystal
sock << "CONNECT #{authority} HTTP/1.1\r\nHost: #{authority}\r\n\r\n"
```

There is no code path anywhere that emits `Proxy-Authorization` (the only hits for that header name are HPACK's static table and the redaction/highlight lists). gori therefore **cannot be used at all** behind an authenticating corporate proxy. This is the one item here where a missing setting is a hard blocker rather than an inconvenience.

**3. No SOCKS support.** `rg -i socks src/` → 0 hits across 623 files. `ssh -D`, Tor, and jump-host workflows are impossible. Burp additionally offers DNS-over-SOCKS.

**4. No TLS client certificates.** `rg -i 'client_cert|pkcs12|p12|pfx' src/` → 0 hits. A target requiring mTLS cannot be tested through gori by any means.

### Proposal

Introduce one `upstream_rules` settings section that supersedes the scalar, keeping `upstream_proxy` as the implicit catch-all rule for backward compatibility:

```json
{
  "upstream_rules": [
    {
      "host_pattern": "*.corp.internal",
      "kind": "http",
      "addr": "proxy.corp.internal:3128",
      "auth": { "scheme": "basic", "username": "…", "password": "…" },
[...]

Issue #403, opened 2026-07-26
fix(proxy): Match&Replace head rewrite desyncs body framing → self-inflicted request smuggling


hahwul:
Found while dogfooding the proxy data path after the recent refactoring sweep. This is self-inflicted request smuggling: an ordinary Match&Replace rule against a **well-formed** client request desyncs the head gori writes from the body it forwards.

## The gap

`src/gori/proxy/conn/client_conn.cr` applies the rewriter to the head that goes on the wire (`sent_head`/`sent_req`, `:210-221`), but computes the body length to stream from the **pre-rewrite** `req`:

```crystal
# :218-219  — the rewritten head is what we send
sent_head = rewritten
sent_req  = Codec::Http1.parse_request_head(rewritten)
...
# :243  — but framing is read off the ORIGINAL req, not sent_req
req_framing, req_len = Codec::Body.request_framing(req)
```

So a rule that touches `Content-Length` / `Transfer-Encoding` makes gori write a head declaring one length while physically forwarding a different number of body bytes. The mirror bug is on the response leg inside `stream_nonhold_response` (`:455-461`, `:622-625`), computed from `response_framing_or_close(resp, …)` on the pre-rewrite `resp`.

## Reproduction (against the built binary)

```
gori run rewriter add --op=set_header --target=request --find=Content-Length --value=4
gori run capture --project P --port 18078
```

Client sends a well-formed request whose 50-byte body begins with 4 bytes then an embedded request:

```
POST /legit HTTP/1.1
Host: 127.0.0.1:19090
Content-Length: 50
Connection: close

AAAAGET /admin-smuggled HTTP/1.1\r\nHost: 127.0.0.1:19090\r\n\r\n
```

The origin parses **two** requests on the keep-alive connection:

```
===== CONNECTION: 2 parsed request-line(s) =====
POST /legit HTTP/1.1
Content-Length: 4
...
AAAAGET /admin-smuggled HTTP/1.1
Host: 127.0.0.1:19090
```

Response leg confirmed the same way (`set_header --target=response --find=Content-Length --value=2` against a 20-byte body → client receives a head declaring 2 bytes but 20 bytes arrive → response splitting downstream). A plain regex `replace` rule triggers it too; an unrelated `add_header` rule does **not** (control: origin gets exactly the declared bytes).

Side effect: gori's own History is corrupted the same way — `gori run show <id>` shows a recorded `Content-Length` head next to a body of a different length.

## Scope

[...]

PR #444, opened 2026-07-27
feat(proxy): upstream connection rules — per-host routing, SOCKS5, proxy authentication

hahwul:
Part of #434 (stays open — see **Still open** below).

> **Stacked on #443.** Base is `hahwul/tls-passthrough`, because both PRs touch `settings/network.cr` and this one builds on `Gori::HostPattern` extracted there. GitHub will retarget to `main` once #443 merges. Review #443 first.

## Problem

The entire outbound path was configured by **one string**, `Settings.upstream_proxy`. Three consequences, all closed here:

1. **No per-destination routing.** "Route `*.corp.internal` through the internal proxy, everything else direct" was inexpressible, and there was no `no_proxy`-style exception list either. The only granularity was global vs. per-project — but a project is usually *one* engagement spanning both internal and external hosts.
2. **No upstream proxy authentication.** `dial_via_proxy` never emitted `Proxy-Authorization` from any code path, so gori simply **could not be used** behind an authenticating corporate proxy. This is the one item where a missing setting was a hard block rather than an inconvenience.
3. **No SOCKS.** `rg -i socks src/` → 0 hits. `ssh -D`, Tor, and jump-host workflows were impossible.

## Change

`upstream_rules` is an **ordered** table, first match wins, sharing the host-pattern dialect with scope host rules (`Gori::HostPattern`, extracted in #443) so a pattern means one thing everywhere. A `direct` rule is the mechanism for carving an exception out of a broader proxy rule above it.

```json
{
  "upstream_rules": [
    { "host": "intranet.corp.internal", "kind": "direct" },
    { "host": "*.corp.internal", "kind": "http", "addr": "proxy.corp.internal:3128",
      "username": "alice", "password_env": "CORP_PROXY_PASS" },
    { "host": "*.onion", "kind": "socks5", "addr": "127.0.0.1:9050" }
  ]
}
```

**One decision point.** `Settings.upstream_route` folds the project pin, the table and the legacy scalar into a single value, so `Upstream.dial` branches once instead of holding three rules that can disagree. Precedence keeps existing installs byte-identical: a project `net.upstream_proxy` pin still wins outright → table → global scalar → direct. An empty table serializes to nothing and behaves exactly as before.

**Credentials are never written to `settings.json`.** A rule stores the username and `password_env` — the *name* of an OS environment variable — and the password is read from the environment at dial time (so `export` takes effect without a restart).

gori's own `env` section is deliberately **not** the mechanism, and this is worth stating because it was the obvious candidate: those variables live in `settings.json` in plaintext, so resolving through them would put the secret in the file by another route and defeat sharing or exporting a config (#439). A `$`-bearing `password_env` is rejected, because that spelling means the operator expected the wrong mechanism.

**CR/LF is stripped from credentials** before they reach the CONNECT head. These values come from `settings.json` and the OS environment, neither validated for wire-safety anywhere else, and a bare CRLF there would inject a header line into gori's own request — the self-inflicted smuggling shape of #403.

**SOCKS5** is RFC 1928 + RFC 1929. A hostname target is sent as `ATYP DOMAIN` so the *proxy* resolves it (`socks5h`), which is what makes Tor and an otherwise unreachable network work at all; gori does not resolve names on the dial path. The reply's `BND.ADDR` is drained by its own address type — otherwise the socket would be handed back mid-reply and the origin stream would start desynced.

## Verification

`crystal spec`: **4456 examples, 0 failures**. ameba on the touched files: 3 findings, identical to the `main` baseline for the same files (the new file contributes none).

23 new specs. Proxy auth and SOCKS5 are asserted **end-to-end against fake servers that report the bytes gori actually emitted**, not by mocking the decision:

- `Proxy-Authorization: Basic …` present with credentials, absent without;
- CR/LF in either credential half does **not** produce an injected header line;
- `ATYP` selection for hostname / IPv4 / bracketed-IPv6 targets, with the IPv6 case checked as 16 raw bytes;
- the RFC 1929 exchange, a credential rejection, and `0xFF` "no acceptable methods" both failing the dial.

Plus rule ordering, the direct-exception carve-out, per-kind default ports (8080 / 1080), scalar fallback, the project pin including its explicit-empty-means-direct case, environment resolution at route time (and the credential vanishing when the variable is unset), and a malformed rule degrading to direct rather than failing every dial.

## Reader's note on one odd line

`socks5_ipv6_bytes` uses `Pointer(LibC::SockaddrIn6)` rather than the `LibC::SockaddrIn6*` sigil. The sigil form placed directly after a multi-line `if … end` in the same method makes the Crystal parser report `can't define def inside def` at the *next* `def`. The `Pointer()` spelling is equivalent and parses; it is not a stylistic preference.

## Still open on #434

- **A TUI editor for the table.** Rules are configured with `gori settings --edit` and documented; a five-field-per-row editor needs its own overlay (a new `OverlayKind`, its dispatch fixtures, and the `on_close` seam), which is a separate change rather than a tail of this one.
- **TLS client certificates**, the fourth gap in #434. They need the `{verify, alpn}` TLS context cache key widened (`upstream.cr`), which #437 also touches — flagged in #434's comment so the two land on one cache key rather than one silently no-op'ing.

## Docs

`reference/config.md` (+ `.ko`): a full `upstream_rules` section with the key table, the credential rationale, the `socks5h` note, and the precedence table.

https://claude.ai/code/session_017X5VpbYNqcLneDkJkYYm1R

### DIFF
--- a/docs/content/reference/config.ko.md
+++ b/docs/content/reference/config.ko.md
@@ -70,6 +70,51 @@ CLI `--listen` / `--port`는 현재 프로세스에 한해서만 이 값들을 �
 
 우회된 호스트는 flow를 남기지 않으므로, gori는 호스트별로 처음 중계할 때 로그에 한 줄을 남깁니다. History에서 빠진 호스트의 이유를 추적할 수 있게 하기 위함입니다. 목록은 Preferences → **Network & Tabs** → **Network** → **TLS passthrough**에서 쉼표로 구분해 편집합니다.
 
+### upstream_rules
+
+목적지별 업스트림 라우팅입니다. `network.upstream_proxy`는 *모든* 트래픽에 대한 단일 주소인 반면, 규칙 테이블은 "`*.corp.internal`은 사내 프록시로, 나머지는 직결"을 표현할 수 있고 자격증명을 실을 수 있으며 SOCKS 프록시에 접근할 수 있습니다.
+
+규칙은 **순서가 있고 첫 일치가 이깁니다**. 구체적인 규칙을 위에 두세요. 편집은 `gori settings --edit`.
+
+```json
+{
+  "upstream_rules": [
+    { "host": "intranet.corp.internal", "kind": "direct" },
+    {
+      "host": "*.corp.internal",
+      "kind": "http",
+      "addr": "proxy.corp.internal:3128",
+      "username": "alice",
+      "password_env": "CORP_PROXY_PASS"
+    },
+    { "host": "*.onion", "kind": "socks5", "addr": "127.0.0.1:9050" }
+  ]
+}
+```
+
+| Key | Type | Description |
+|-----|------|-------------|
+| `host` | string | 호스트 패턴. 스코프 `host` 룰과 같은 문법 — `corp.internal`은 해당 호스트와 서브도메인, `*.corp.internal`은 글롭, `*`는 catch-all. 대소문자 무관 |
+| `kind` | string | `direct`, `http`, `socks5`. 알 수 없는 kind는 규칙을 버립니다(`direct`로 취급하면 의도한 프록시를 조용히 비활성화하게 되므로) |
+| `addr` | string | 프록시 `host:port`. 포트 기본값은 `http`가 `8080`, `socks5`가 `1080`. `direct`에는 없어야 합니다 |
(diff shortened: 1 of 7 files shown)


hahwul:
Superseded by the reopened PR below — GitHub auto-closed this one when its base branch `hahwul/tls-passthrough` was deleted after #443 merged, and a closed PR can be neither reopened nor retargeted to `main`. Same branch and commit, rebased onto `main`.

Issue #320, opened 2026-07-22
Feature Request: Support user-defined custom Active Rules in Probe scanner


hahwul:
### Feature Description
Currently, Probe scanner's `CustomRule` (`src/gori/probe/custom_rule.cr`) only supports passive string/regex matching rules. All active scan checks (`Active::Rule`) are hardcoded built-in Crystal classes (`ReflectedParam`, `CorsReflection`, `ForbiddenBypass`, `NginxAliasTraversal`, `BackslashPowered`).

### Motivation / Use Case
It would be highly valuable if users could define custom Active Rules (e.g. via YAML/JSON config or UI/DB) to send custom probe requests (with specific payload injections) and analyze response variations.

### Suggested Implementation
- Extend the Probe custom rule schema to support Active Rule definitions (e.g., target parameter/header, probe payloads, response comparison criteria).
- Allow users to load and execute custom active scan checks in TUI/CLI.

---
요약: Probe 스캐너에서 패시브 규칙 외에 사용자 정의 프롭 요청 및 응답 비교(Active Rule)를 지원하는 커스텀 Active Rule 기능 요청.

hahwul:
Not planning this after all — a product-direction call.

A user-defined active rule is a single payload + pattern + control diff. What that can catch is mostly already covered by the built-in active rules (traversal, CRLF, reflected params, nginx alias), the batch-scan use case it competes in is nuclei's, and the flexible version of "inject a payload and judge the response" already exists through the MCP seam (`send_request` + AI judgement) with none of the schema/CRUD weight.

Most of all it pulls against gori's direction — reject the scanner, wedge on the manual loop + MCP. Prototyped the differential engine in #482 (closed); it's in git history if this ever gets reconsidered.

PR #482, opened 2026-07-29
feat(probe): custom ACTIVE rules — the differential core (#320, 1/N)

hahwul:
Closes the first slice of #320. The Probe custom-rule system is passive only today — a string/regex matched against captured bytes, no request sent. #320 asks for user-defined **active** rules: send a payload, judge the response. This PR is the engine for that, as a **pure, fully-specced core**. Persistence (schema V5 + model + global settings) and the TUI/CLI/MCP CRUD surfaces are separate follow-ups so this stays reviewable.

## What's here

`CustomActiveRule` — a data record (id, inject location, payload, confirming pattern, region, severity, scope, enabled). `to_rule` wraps it in an `Active::Rule` the existing pipeline drives with **zero special-casing**: `Active.analyze` gains a `custom:` parameter and chains the enabled rules after the built-ins through the same send → confirm loop.

## The design: differential by construction

A rule that only asked *"does the response match the pattern?"* would fire on every page that already contained the pattern's text — the exact false-positive class #478 removed from the built-in bypass rules. So the adapter always sends **two** requests:

- **probe** — the payload injected
- **control** — the request verbatim

and reports only when the pattern is **present in the probe AND absent from the control**. The payload produced the match, not the page. A control that failed to send is no attribution, so it **refuses** rather than falling back to the probe alone — the same discipline as the four rules in #478.

## Injection & gating

Three locations, the ones a user rule needs: append the payload to every query value, set a named request header, or append to the body.

- query / header — SAFE_METHODS-gated by default (a user payload can be anything, so re-sending it on a POST/PUT/DELETE stays an explicit `allow_unsafe` opt-in)
- body — needs a body, and the opt-in for a non-GET

Response matching reuses the passive rule's byte-safety exactly: heads/bodies scrubbed before the regex, a bad user pattern degrades to no-match rather than dropping the probe.

Header and payload bytes go out **verbatim** — a CR/LF the operator typed is not stripped, per the project's provenance rule (operator bytes are byte-exact; a `foo\r\nX-Injected: 1` payload is the operator deliberately testing header injection, not a mistake to sanitize).

## Verification

- **15 new specs**: validation, the probe/control shape for each injection location, the present-in-probe / absent-in-control differential (with its no-fire and refuse cases), regex + region matching, method gating, `dedup_key`/`plan` equivalence, and the `Active.analyze` wiring skipping a disabled rule.
- `crystal spec` — **4815 examples, 0 failures**
- `crystal tool format --check`, `crystal build --no-codegen`, ameba on the touched files — all clean

## Not in this PR (follow-ups)

Schema V5 + the `ProbeCustomActiveRule` model, global settings storage, and the TUI/CLI/MCP CRUD. The engine is exercised end-to-end by the specs, so it is **not dead code** while those land.

## Design note for review

I kept the confirmation to a single mode (probe-present / control-absent) rather than exposing a `confirm:` knob — it is the one mode that is FP-safe, and adding others later is additive. Happy to widen it if you'd rather the record carry the mode from the start.

### DIFF
--- a/spec/probe/active/custom_active_rule_spec.cr
+++ b/spec/probe/active/custom_active_rule_spec.cr
@@ -1,0 +1,247 @@
+require "../../spec_helper"
+
+# --- file-local harness ----------------------------------------------------------------------
+
+private def with_store(&)
+  path = File.tempname("gori-customactive", ".db")
+  store = Gori::Store.open(path)
+  begin
+    yield store
+  ensure
+    store.close
+    File.delete?(path)
+    File.delete?("#{path}-wal")
+    File.delete?("#{path}-shm")
+  end
+end
+
+private def capture_flow(store, *, target = "/s?q=hi", method = "GET", status = 200,
+                         req_headers = "", req_body : String? = nil,
+                         host = "acme.test") : Gori::Store::FlowDetail
+  head = String.build do |io|
+    io << method << " " << target << " HTTP/1.1\r\nHost: " << host << "\r\n" << req_headers << "\r\n"
+  end
+  req = Gori::Store::CapturedRequest.new(
+    created_at: 1_000_i64, scheme: "https", host: host, port: 443,
+    method: method, target: target, http_version: "HTTP/1.1",
+    head: head.to_slice, body: req_body.try(&.to_slice))
+  id = store.insert_flow(req)
+  store.update_response(Gori::Store::CapturedResponse.new(
+    flow_id: id, status: status, head: "HTTP/1.1 #{status} X\r\n\r\n".to_slice, body: nil,
+    reason: "OK", content_type: nil, duration_us: 1_i64))
+  store.get_flow(id).not_nil!
+end
+
+# A backend that reflects the DIFFERENTIAL: a request carrying `needle` gets a response body
+# containing `marker`, everything else gets a clean body. Keyed on the request content rather than
+# on send order, so the built-in rules (which run first and send their own probes) can't consume
+# the responses meant for the custom rule's probe/control pair.
+private class DifferentialBackend < Gori::Fuzz::Backend
+  getter origin : Gori::Fuzz::Origin
+  getter sent = 0
+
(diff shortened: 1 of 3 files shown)


hahwul:
Design review resolved — updated to match the two decisions:

1. **Confirmation mode: single.** Kept the one FP-safe mode (present-in-probe, absent-in-control). No `confirm:` field, so the storage PR won't land a column that has a single value.
2. **Injection: five locations.** Added **cookie** and **path** to query/header/body.
   - `cookie` — appends the payload to every value in the request Cookie header, one line rewritten in place. **Raw**, not URL-encoded: a cookie value isn't URL-decoded server-side the way a query value is. Gated on a non-blank Cookie header.
   - `path` — appends to the URL path, query preserved (`/a/b?q=1` → `/a/b<payload>?q=1`). Also **raw**: path injection is for traversal / normalization tricks, and encoding the `/` and `.` would defeat it. Always applicable.

Specs 15 → 18 (cookie rewrite, path+query preservation, the two new gates). Suite 4818 green; format/build/ameba clean.

Design doc for reference: https://claude.ai/code/artifact/e402a34a-6e0f-44e6-8021-c928b7892602

hahwul:
Closing — product-direction call, not a code problem.

The differential engine works, but the primitive is a single payload + single pattern + control diff, and on reflection that is:

- **redundant with the built-ins** for most of what it can catch (traversal file contents, error signatures, reflected markers already belong to `lfi_param_traversal` / `crlf_injection` / `reflected_param` / `nginx_alias_traversal`),
- **a thin subset of nuclei** for the batch-scan use case it competes in, and
- **already covered, more flexibly, by the MCP seam** — `send_request` + AI judgement does payload-inject → control re-send → verdict with none of the schema/CRUD weight.

Most of all it pulls against gori's stated strategy: reject the scanner, wedge on the manual loop + MCP. A user-defined batch-scan rule is the opposite direction. Not worth the schema V5 + three CRUD surfaces for the narrow "org-specific signature" niche that's left.

Engine is preserved in git history if the calculus ever changes.

PR #665, opened 2026-08-12
Performance: transport, render, store — and eight proposals the measurements rejected

hahwul:
Nineteen commits across the testing engines, the TUI render path, the store, and
two crash classes. Every change here has a number attached, and so does every
change that is **not** here: eight proposals were built, measured, and reverted,
which is most of what this branch learned.

## Crashes

**Sitemap.** `post_order` landed with a comment naming the SIGSEGV deep paths
caused, but only two of the walks were converted. Ten more were still recursive,
including the three in `gori run sitemap`'s output layer — the second surface
that comment names. Measured in a debug build: only `keep_for_tags?` (40k
segments) and `SitemapView#collect` (80k) actually overflow; the rest are
converted anyway because it is cheap and removes the question of which is safe.

The real ceiling was elsewhere and nobody had looked. `Sitemap.build` is
**quadratic in path depth** — every node stores its full path from the root — so
one captured request costs 1.6 GB at 20k segments, 6.6 GB at 40k, and is
OOM-killed at 160k, with no walker running at all. `MAX_DEPTH = 128` cuts a
target and flags the node, which the TUI label, the CLI text tree and the CLI
JSON all say out loud. 20k now costs 6.9 MB.

**Supervision.** The TUI run loop had `ensure` and no `rescue`, and `CLI.run`
catches only `Gori::Error` — so one `IndexError` on a render tick ended a session
holding unsaved Repeater/Fuzzer buffers and a held intercept queue. Now absorbed
per tick with a three-in-ten-seconds breaker. The five probe analyzer fibers run
under `supervise`; losing `catch_up_loop` silently meant captured traffic stopped
being scanned for the rest of the session. Checked before patching: all four MCP
job runners already had `rescue` + `ensure` — that part of the report was wrong.

Also the last PCRE2 hole in the editor: `env_value_preview` ran a regex over a
binding value that came off the wire, and PCRE2 raises on invalid UTF-8.

## Speed

| what | before | after |
| --- | --- | --- |
| Editor keystroke, 512 KB buffer | 5.53 ms | **0.098 ms** |
| `Outbound.request_target`, 50 KB POST | 17.0 µs / 50.4 kB | **173 ns / 256 B** |
| Sequencer, 25 samples | 25 handshakes | **1** |
| Open-findings count, 250k rows | 146.8 ms | **6.28 ms** |
| Unindexed History filter, 40k rows | 15.2 ms | **7.8 ms** |
| Miner probe+decide | 54.6 µs | **37.2 µs** |
| Probe triage refresh, 250k rows | 137.5 ms | **107.3 ms** |
| Fuzzer sorted view, live run | 615 µs + 2.54 MB per frame | throttled to 4/s |

The editor now styles the rows on screen rather than the whole buffer on every
keystroke, using the `Highlight::Windowed` machinery Intercept has had all along.
Because it changes rendering, the gate is that it paints the *same* thing: the
windowed and eager paths are asserted equal line for line, span for span, on
text, colour and attribute.

`Fuzz::Sender`'s comment said the "one-shot senders (Repeater minimize, Probe
active) have nothing to amortise" and every caller believed it. None is one-shot.
Turning keep-alive on was the easy half; the other half is that whoever turns it
on owns closing the pool, so three engines grew an `ensure`. `gori run sequence`
gets `--no-keep-alive`, because that tool's output is a statistical claim about
how an origin generates tokens and the transport is therefore not neutral.

Counting open findings moved into SQL. Five sites — the TUI's dismiss-by-host,
both MCP dismiss verbs, both CLI ones — loaded every `probe_issues` row to
produce one integer, and that integer is what the operator is told ("dismissed
N"). The planned fix for this table was a `LIMIT`; it would have made all five
under-report.

## Scale

`cache_size` at 64 MiB (the default is 2 MiB **per connection**) and a 256 MiB
`mmap_size`. Two triage indexes, because `probe_issues` and `issues` were both
read whole and sorted with nothing backing the sort. `events` gets the only
retention path it never had — scoping that found `oast_callbacks`, `fuzz_runs`
and `intercept_commands` are all already cleaned up elsewhere, so "five uncapped
tables" was one.

## Built, measured, reverted

This is the part worth reading.

- **One-pass frame fill** — slower and allocated more (284 µs / 260 kB →
  332 µs / 416 kB).
- **`flows` covering index** — the plan's "biggest scale win". Genuinely used
  (`SCAN flows USING COVERING INDEX`) and still not worth it: 7.5 → 6.5 ms on
  the worst case, for 3% off sustained INSERT and ~20 MB per 100k rows. The
  `cache_size` change above had already taken that win; the premise assumed the
  2 MiB default. The two filters that are actually slow (`header:` 164 ms,
  `body:` LIKE 452 ms) scan the BLOBs and no projection index reaches them.
- **Discover pool LRU** — three defects: `close_all` runs SSL_shutdown, a write
  and so a yield point, landing one exactly where `pool_for` documents that none
  may be; eviction can only drain the idle list, orphaning checked-out sockets;
  and interleaved origins make it thrash worse than the cliff it replaced.
- **Probe Active sender cache** — no change at all (9 requests over 5
  connections either way). Most Active probes are unpoolable by framing, and the
  reuse that happens is within a task, which the per-task sender already gets.
- **`max_idle_pool_size=4`** — left `#close` unable to checkpoint: the db file
  stayed at 4096 bytes with a 1 MB `-wal` beside it.
- **History preview throttle** — 60 µs against a ~300 µs frame, and it broke a
  documented contract that a pending flow keeps refreshing.
- **`build_dist`** — already fixed; the bench compares OLD against NEW and NEW
  is what ships.
- **Review finding #8** (Probe Active pooling biasing the smuggling rule's
  timing verdict) — the mechanism does not hold. `ConnPool` refuses to park a
  socket that carried an ambiguous framing, which is what every one of those
  probes is, so they still dial fresh and stay comparable to the baseline. The
  real residue is smaller: the second benign baseline is now the faster of the
  two, so the stability twin no longer cross-checks the first. Pinned by spec,
  because loosening the pool's predicate would break the comparability silently.

## Also

17 of the 43 `bench/` harnesses did not compile. `just check` only *formats*
bench/, and CI runs build, spec and docker, so they had rotted silently — the
harnesses AGENTS.md points at for "measure, don't guess". Fifteen required a
subtree that stopped being self-contained; two had drifted off
`FlowSink#on_ws_message` after it grew a parameter, whose comment claimed the
default spared implementors (it spares callers). `just benchmark-check` now
type-checks all 43.

## Verification

`crystal spec` 8319 examples, 0 failures. Format clean, no new ameba findings.
Rebased on main and the merge result builds; all 43 harnesses build. Migration
verified on a real v14 database (→ v15, data intact) and a v99 one is refused
without being touched. Smoke-tested in a pty: tab switching, a live capture
through the proxy, the sitemap truncation marker end to end in all three CLI
formats, and an edit-and-resend round trip in the Repeater — `gori.log` empty at
the end.

## Not done

Each needs a contract decided before code: `S3` listing LIMIT (truncation has to
be reported on three surfaces), `S4` import chunking (breaks a documented
atomicity guarantee), `S5` beyond `events`, `T5` incremental sitemap reload,
`X3` ConnPool checkout (needs `SSL_pending` and the pool holding the plaintext
fd), `X4` retry pacing (making a failing origin retry sooner is not obviously
right), `X9` h2 connection reuse (`h2_engine.cr` is hardcoded to stream id 1 in
seven places).

`Store#close` depends on `max_idle_pool_size` being 1 to get SQLite's
last-connection checkpoint; that is documented in place, not fixed. `bench/` is
still not wired into CI.


hahwul:
Reviewed and acted on. Five of the nine were defects this branch introduced.

**Fixed** (`f2acebff`):

- **#2/#3/#4 — the discover LRU is reverted.** All three are right and they
  compound: `close_all` runs SSL_shutdown, which is a write and therefore a
  yield point, landing one exactly where `pool_for` documents that none may be;
  eviction can only drain the idle list, so a checked-out socket is orphaned
  past `Sender#close`; and interleaved origins make it thrash worse than the
  cliff it replaced. The spec I rewrote exercised only sequential per-origin
  bursts — the one pattern LRU handles perfectly, which is why it looked fine.
  Original behaviour and spec restored, with the reasoning recorded so the next
  attempt starts from "ConnPool must report in-flight checkouts".
- **#5 — MCP discover now flushes on time too.** Confirmed the TUI twin batches
  by tick (`flush_persist if applied`, ~50 ms), not by count. Under 64 findings
  they were invisible to `list_sitemap`/`get_flow` for the whole run.
- **#6 — one-entry memo over `line_at`.** Confirmed and measured: 4 KB lines
  filling a 40-row wrapped pane cost **0.35 ms/frame** without it, **0.055 ms**
  with. Added as a bench case so it cannot regress silently again.
- **#1 — the breaker's toast goes through `status()`.** Exactly right: without
  `@toast_at` the companion bubble won the recency comparison and the only
  operator-visible sign the breaker fired was swallowed.
- **#9 — `max_pool_size=8`.** 64 MiB per connection against an unbounded pool
  was the wrong pairing.
- Also corrected the `each_line` comment; it does chomp `\r\n`, and the
  equivalence rests on `strip`/`split` treating the CR as whitespace, not on the
  two producing the same string.

**Left as follow-ups, recorded in the commit message:**

- **#7 — no `--no-keep-alive` on the three senders.** Fair, and the Sequencer is
  the sharp case: its output is a statistical claim about token generation, so
  an origin with connection-bound issuance now yields a sample the operator
  cannot re-take. Plumbing the flag is three surfaces and belongs in its own
  change.
- **#8 — the smuggling rule's cold-baseline bias.** Real and one-directional,
  but well inside `DELAY_THRESHOLD = 4s`; worth fixing where the baseline is
  chosen rather than by reaching back into the transport.

Verification after the fixes: `crystal spec` 8309 examples, 0 failures; format
clean; all 43 bench harnesses build; keystroke win intact (512 KB buffer
5.53 ms → 0.086 ms).

Issue #739, opened 2026-08-17
docs: the proxy guide's protobuf and WebSocket-compression claims are both stale


hahwul:
Two claims in the Protocol Support section of the proxy guide are no longer true, and both understate what gori can do.

## 1. The gRPC row says protobuf is raw bytes — the decoder shipped in #505

`docs/content/guide/proxy.md:128` and `docs/content/guide/proxy.ko.md:125`:

> | **gRPC** | Framed over HTTP/2 with status trailers; protobuf shown as raw bytes (no `.proto` schema) |

The schema-less wire-format decoder landed in #505 (closing #496): `src/gori/protobuf.cr`, emitted by the CLI at `src/gori/cli/run/history.cr:483` and by MCP at `src/gori/mcp/serialize.cr:427` as `grpc_messages[].protobuf`.

Worse than a wrong line: grepping all of `docs/` finds **no page anywhere that mentions the decoder**. `docs/content/guide/mcp.md` and the CLI docs describe `reframe_grpc` but never the `protobuf` field. A shipped feature is entirely invisible.

Fix: correct both rows, and document the `grpc_messages[].protobuf` tree in the CLI/MCP pages, including the "compressed payload — not decoded" note and the trailer-frame carve-out.

(Note: the TUI genuinely does still show hex. That is tracked separately as a parity gap, and the doc line should say which surfaces decode.)

## 2. TLS passthrough is offered as the only way to keep a WebSocket byte-exact — there is a better one

`docs/content/guide/proxy.md:135` and `docs/content/guide/proxy.ko.md:132` end the permessage-deflate explanation with:

> If you need a particular host's sockets relayed exactly as they are, put it under TLS passthrough, which leaves the connection alone and captures nothing for it.

But the strip is deliberately ordered **before** Match & Replace precisely so a rule can put the offer back — `src/gori/proxy/conn/client_conn.cr:488-501`: *"an offer hidden from this scan … operator bytes win here"*. When a rule restores the header, `settle_ws_extensions` sees `carries_extensions?(sent_head) == true` and relays the origin's acceptance untouched (`client_conn.cr:1337-1342`; `spec/proxy/ws_spec.cr:1338-1372` asserts `# P7: untouched`).

For most operators that is the strictly better escape hatch: a byte-exact wire **and** a flow that is still captured, with a `[gori]` notice explaining that the frames are encoded bytes — versus passthrough, which captures nothing at all. Two sentences, next to the existing paragraph.

---

[...]

Issue #738, opened 2026-08-17
feat(proxy): the proxy listener cannot answer a client configured with an HTTPS proxy


hahwul:
A client configured with `https://proxy:port` — Firefox's "HTTPS Proxy" field, or Chrome via a PAC file returning `HTTPS ...` — cannot use gori. The proxy listener speaks cleartext only.

`LISTENER_MODES = ["proxy", "transparent", "reverse"]` (`src/gori/settings/listeners.cr:47`), and proxy mode goes straight to a cleartext `ClientConn` at `src/gori/proxy/server.cr:184-186` with no ClientHello branch. The other two modes both sniff `0x16` first (`server.cr:234-237`, `server.cr:280-289`); the proxy mode does not.

So a TLS ClientHello arriving on the proxy port is parsed as an HTTP/1.1 request line, fails, and the connection dies.

## Worth noting, low priority

The pentest impact is small — nearly every client that supports an HTTPS proxy also supports a plain HTTP one, and the connection to `127.0.0.1` is not the leg that needs confidentiality. It is adjacent to the already-open #504 ("reach clients that ignore proxy settings") but distinct: #504 is about clients that *ignore* proxy configuration, this is about clients configured for a proxy gori cannot answer.

Filing it so the listener-mode matrix is on the record. Close as `wontfix` if the audience is judged too small.

---

한줄 요약: proxy 모드 리스너는 평문 전용이라 `https://proxy:port`로 설정된 클라이언트가 gori에 붙을 수 없습니다(reverse·transparent는 `0x16`을 sniff하는데 proxy만 안 함).

hahwul:
Closing as wontfix, as this issue itself proposed.

The gap is real and the code facts stand: `LISTENER_MODES = ["proxy", "transparent", "reverse"]` (`src/gori/settings/listeners.cr:47`), and proxy mode goes straight to a cleartext `ClientConn` (`src/gori/proxy/server.cr`) with no ClientHello branch, unlike the other two which both sniff `0x16`. A client configured with `https://proxy:port` cannot use gori.

It is not worth building, for two reasons that compound:

**The audience is small.** Effectively every client that supports an HTTPS proxy also supports a plain HTTP one, and the leg in question is to `127.0.0.1` — the one hop that does not need confidentiality. Nothing in the workflows gori is built for is blocked by this.

[...]

Issue #737, opened 2026-08-17
feat(proxy): serve h2c prior-knowledge on the reverse and transparent listeners


hahwul:
gori can **send** h2c prior-knowledge but cannot **receive** it, and the two listener modes where receiving it is well-defined are exactly the two that refuse it.

- Sending: `src/gori/discover/plan.cr:79` and `src/gori/repeater/h2_engine.cr:414-421` both do h2c prior-knowledge to an origin.
- Receiving: the reverse listener (`src/gori/proxy/server.cr:232-242`) and the transparent listener (`src/gori/proxy/server.cr:279-295`) route on `0x16` only. There is no `0x50` branch anywhere, so `PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n` falls through to `ClientConn`, which rejects it at `src/gori/proxy/conn/client_conn.cr:271-275` and closes.

## Why these two listeners specifically

This is not the same request as "serve h2c on the forward-proxy port". On the forward-proxy port, refusing is correct: RFC 7540 §3.4 prior knowledge is defined for a direct origin connection, and a forward proxy would need `:authority`-based routing that the relay explicitly disclaims (`src/gori/proxy/conn/client_conn.cr:1256`). Two clients confirm the point empirically — `curl --http2-prior-knowledge` through a proxy prints `Ignoring HTTP/2 prior knowledge due to proxy` and downgrades, and grpc-go with `HTTP_PROXY` set issues a `CONNECT`, which already works.

On a **reverse** listener gori *is* the origin (fixed host/port), and on a **transparent** listener it is in-path with the target from `SO_ORIGINAL_DST`. In both cases `:authority` routing is unnecessary and prior knowledge is exactly as well-defined as it is for the senders above. `curl --http2-prior-knowledge http://host/` with no proxy sends the preface, and gori closes the connection.

## Use case

Cleartext gRPC between pods: gori as a reverse listener in front of an h2c service, or transparently intercepting via pf/iptables. Service-mesh and internal-API work generally, where TLS terminates at the edge and the interesting hop is h2c.

## Scope

Add a `0x50` branch to the two listeners, routing to the existing `intercept_h2c` (`src/gori/proxy/conn/client_conn.cr:1281-1302`) — the machinery, including the rewriter/interceptor/extractor wiring from #549, is already there and both listeners already have the host/port it needs.

Niche audience, so low priority; filing it so the send/receive asymmetry is on the record rather than rediscovered.

---

[...]

Issue #736, opened 2026-08-17
fix(proxy): a non-WebSocket 101 upgrade is blind-tunnelled with no notice on the flow


hahwul:
When a 101 response is not a WebSocket upgrade, gori pipes the rest of the connection raw and says nothing about it:

```crystal
# src/gori/proxy/conn/client_conn.cr:824-840
if websocket_upgrade?(resp)
  WS::Relay.run(@io, upstream, flow_id, @sink, @rewriter, ws_ctx, @interceptor, notice: ws_notice)
else
  Pump.blind_tunnel(@io, upstream)
end
```

The tunneling itself is fine and documented as intentional (`src/gori/proxy/pump.cr:1-4`, `client_conn.cr:812-815`). The upgrade request and the `101 Switching Protocols` head are already in History as a `Complete` flow — `stream_nonhold_response` records them before the branch at `:807` — so this is *not* zero capture. Only the post-upgrade byte stream is opaque.

**The gap is the asymmetry.** The WebSocket branch carries a per-flow `notice` (`client_conn.cr:749-752`, `:837`; `src/gori/proxy/ws/relay.cr:280`) that tells the operator what gori did to the socket. The non-WebSocket branch carries nothing — not a flow note, not a `gori.log` line. "gori saw this upgrade and deliberately did not decode it" is never on the record.

## Who actually hits it

Live protocols, not dead ones:

- **kubectl `exec` / `attach` / `port-forward`** — `Upgrade: SPDY/3.1`
- **Docker Engine API `attach` / `exec`** — `Upgrade: tcp`

(RFC 2817 `Upgrade: TLS` and h2c are both effectively dead and not worth considering.)

## Scope

A `gori.log` line plus a flow-level note naming the `Upgrade:` token and the bytes relayed in each direction, mirroring the WS `notice` seam.

Explicitly **not** in scope: decoding SPDY or `Upgrade: tcp`. That would need per-protocol decoders and is out of proportion to the payoff.

---

한줄 요약: WebSocket이 아닌 101 업그레이드(kubectl exec의 SPDY, Docker API의 `Upgrade: tcp`)는 blind 터널로 흘려보내면서 WS 분기와 달리 flow note도 로그도 남기지 않습니다.

hahwul:
Fixed by #741 (`5a993b96`), verified against `main` today.

[...]

PR #741, opened 2026-08-17
Protocol support gaps: nine fixes from an audit sweep

### LINKED ISSUE #728: fix(proxy): `Expect: 100-continue` deadlocks the HTTP/1.1 path
A client that sends `Expect: 100-continue` and waits for the interim response — as the RFC tells it to — never gets one through gori. Client, proxy and origin then wait on each other until a timeout breaks the tie.

## Reproduced

Fresh build of `d1571a9f`, a Python origin on `:18901` that answers `100 Continue` from `handle_expect_100`, `gori run capture -p 18902` in front of it. The client writes the head only and then blocks on `recv`, which is what a conformant client does:

| | time to `100 Continue` |
|---|---|
| client → origin (control) | `0.00s` |
| client → gori → origin | **never arrives; client gives up at 8.00s** |

The origin log proves it is not the origin's fault — it sends the 100 and then blocks reading a body that never comes:

```
[origin] 0.67s sending 100 Continue
[origin] 0.67s reading body of 2048
```

## Why

All three request paths write the head upstream and then immediately block copying the client's body, so nothing ever reads the origin's interim response:

- `src/gori/proxy/conn/client_conn.cr:446-451` — non-hold streaming path:
  ```crystal
  upstream, reused, sent = acquire_and_send(host, port, retryable) do |up|
    up.write(sent_head)
    req_complete = Codec::Body.stream(@io, up, req_framing, req_len, req_capture, copy_buf)
    up.flush
    true
  end
  ```
  `handle_response` — which contains the interim handling — is not reached until `:467`.
- `src/gori/proxy/conn/client_conn.cr:509` — the intercept-hold path calls `Codec::Body.read_complete(@io, …)` *before* the head goes upstream at `:542`, so the origin is not even given the chance to send a 100.
- `src/gori/proxy/conn/client_conn.cr:652-654` — the body-rewrite path has the same shape.

Three-way deadlock: the client waits for the 100, gori waits for the body, the origin waits for the body.

## The machinery is already correct — it just runs too late

`skip_interim_responses` (`src/gori/proxy/conn/client_conn.cr:869-925`, `:1360-1371`) is complete and careful: it forwards the 1xx, refuses a body-declaring 1xx, caps at `MAX_INTERIM`, and withholds 1xx from an HTTP/1.0 client at `:899`. This is purely an ordering bug, not a missing feature.

Both Repeater engines already handle this case explicitly — `src/gori/repeater/engine.cr:225` and `src/gori/repeater/h2_engine.cr:528` both name `Expect: 100-continue`. The proxy path is the odd one out.

**HTTP/2 is unaffected.** `src/gori/proxy/h2/relay.cr:49-50` spawns two independent pump fibers, so an interim HEADERS frame reaches the client regardless of DATA progress.

## Impact

curl adds `Expect: 100-continue` on its own for any body over ~1 KB, so large uploads through gori stall until curl's 1-second self-recovery timer fires. A client that waits properly (legacy .NET `Expect100Continue`, many Java/SOAP and mobile stacks) hangs for the full 30 s `CLIENT_IO_TIMEOUT` (`src/gori/proxy/socket_tuning.cr:34`) and is then handed a 502. To an operator this reads as "gori breaks file uploads".

## Fix shape

After writing the head, if the request carries `Expect: 100-continue`, read the upstream for an interim response and relay it before blocking on the client's body — or run the body copy on its own fiber so neither direction can starve the other.

## Spec gap

`spec/proxy/server_spec.cr:357-400` only covers an origin volunteering a 100 on a bodyless GET. No spec sends `Expect:` from the client, which is why this survived.

---

한줄 요약: 클라이언트가 `Expect: 100-continue`를 보내고 기다리면 gori가 오리진의 `100 Continue`를 중계하지 않아 3자 교착이 발생하며, 큰 업로드가 멈춥니다.

hahwul:
Nine issues from a protocol-support audit, each on its own commit. Closes #728, #730, #731, #732, #733, #734, #735, #736, #737.

The audit started from "which protocols can gori not handle". Most of the answers turned out to be settled decisions or already-closed work — HTTP/3 (#719 decided the mitigation, not the codec), the h2↔h1 bridge (#492 step 5, closed with flow-control measurements), permessage-deflate decode (#518), `.proto` schema support (#505 shipped the schema-less decoder), SOCKS inbound (#504). What is here is what survived that filter.

## Commits, in landing order

| commit | issue | what |
|---|---|---|
| `02794347` | #732 | Firefox launched from `^P` silently proxied no localhost traffic. `no_proxies_on = ""` cannot reach it — `CanUseProxy` refuses loopback *before* the bypass list. |
| `7a75789a` | #735 | `Transfer-Encoding: gzip` was neither inflated nor labelled — silently worse than an unknown `Content-Encoding`, which at least says "decode unsupported". |
| `d2da4d95` | #730 | The h2 downgrade gate runs once per CONNECT; the rule table is live-mutable. A body rule enabled on an already-open h2 connection applied to nothing and said nothing. |
| `8f80345a` | #728 | `Expect: 100-continue` deadlocked all three request paths. Highest-impact item here. |
| `443146b9` | #734 | The schema-less protobuf decoder (#505) reached CLI and MCP but not the TUI — headless was strictly richer than the TUI for gRPC. |
| `d4d3ee48` | #736 | A non-WebSocket 101 (kubectl `SPDY/3.1`, Docker `Upgrade: tcp`) was blind-tunnelled with no note, while the WebSocket branch has carried one all along. |
| `2cb35000` | #737 | h2c prior knowledge is served on the reverse and transparent listeners. gori could already *send* it and not *receive* it. |
| `5fb83066` | #731 | The h2c-in-CONNECT sandbox refusal was stale, and made `H2::StreamGate`'s per-stream gate unreachable dead code. |
| `87dbe278` | #733 | WebSocket over HTTP/2 (RFC 8441 extended CONNECT) is now captured, using the existing WS codec over the stream's DATA frames. |
| `58e1870a` | — | Integration: #737 and #731 each compiled alone and did not compile together. See below. |

## Where to look hardest

**`8f80345a` (#728) — the biggest behavioural change.** Between writing the request head and pumping the client's body, the streaming path now settles the expectation four ways: relay the origin's interim verbatim; relay a *final* answer (417/401/30x) and send **no** body; stop on a dead origin; or, after 1 s, write the `100 Continue` itself. The three buffering paths (intercept hold, body rewrite, and `serve_short_circuit`, which the issue did not list but has the identical shape) generate the interim themselves because they must hold the whole body before anything is dialled.

Worth a reviewer's attention:
- retry safety rests on `retryable_request?` returning false for every body-declaring request, so `acquire_and_send` structurally cannot re-run the block. If that invariant ever changes, this double-sends.
- the final-response branch forces the client connection closed, because the un-sent body is still owed on the client socket. Removing that forced close reproduces the desync on demand: the follow-up request line is eaten as the body (`Expected: 0, got: 45`).
- gori issuing its own `100` is a real semantic addition, not just plumbing. The origin no longer gets to refuse a body it never saw, on the buffering paths.

**`87dbe278` (#733) — new subsystem, and one lifecycle change.** `emit_ready` used to delete a stream on `resp.ended && flow_id`, safe only because `flow_id` implied the request half had closed. An extended CONNECT's request half stays open for the socket's life and the flow must be projected at the request head, so that implication no longer holds; an origin closing first would have dropped the client's half of the §7.1.1 closing handshake. Please check that reasoning independently.

**`5fb83066` (#731) — removes a security gate.** The argument is that `H2::StreamGate` has carried a fail-closed per-stream sandbox since #492 step 4, and #549 wired the interceptor into `intercept_h2c`, so the blanket refusal sat in front of a gate that was already there. The spec that carries this proves it both ways: stub `sandbox_refuses_locked` and the out-of-scope stream reaches the origin (`Expected: [3], got: [1, 3]`).

## Two things the audit got wrong, corrected in the work

- #737 as filed proposed a `0x50` first-byte branch. On a **listener** that byte also opens `POST`/`PUT`/`PATCH`, so it would have diverted every form submission into the h2 relay. The implementation compares the real 24-octet preface with a four-byte floor. Recorded in a comment on the issue.
- #736 as filed said "zero capture". The upgrade request and the 101 head were already recorded; only the post-upgrade byte stream is opaque.

## The integration commit, and a verification gap it exposed

#737 landed its listener entry point as a `ClientConn` reopen in `server.cr` calling `h2c_unservable?`; #731 replaced that predicate with `h2c_refusal`. Each branch was green in isolation and the merge did not compile.

It did not fail loudly either: `crystal build --no-codegen src/gori.cr` accepted it, because `src/gori.cr` is the library root ("no top-level execution so it can be required by specs" — its own first line) and Crystal does not typecheck a method body nothing calls. **`crystal build --no-codegen src/main.cr` is the check that names the break.** Worth knowing for CI and for anyone verifying a branch locally.

## Deliberately not here

- HTTP/3 / QUIC. There is no UDP socket anywhere and that is the decision from #719, not an oversight.
- Body rewriting on HTTP/2 (#492 step 5) and permessage-deflate decode (#518) — both closed with reasons that still hold.
- `#738` (an HTTPS-proxy listener) — it needs a leaf-certificate story that has not been designed.

## Known follow-ups

- **The #733 transcript is not visible yet.** A WS-over-h2 flow is `CONNECT → 200`, and eight readers gate the transcript on `status == 101` (`history_view.cr:1025`, `repeater_controller.cr:1394`, `mcp/tools/flows.cr:69`, `mcp/tools/repeater.cr:172`, `cli/run/history.cr:245`, `cli/run/repeater.cr:294` and `:1238`, `export/har.cr:167`). The honest predicate is "does this flow have `ws_messages` rows"; `Store#count_ws_messages` already exists. The three repeater-seed sites need a decision rather than a rename, because `WsEngine` replays over an h1 handshake.
- `docs/content/guide/proxy.md`'s protocol table still needs the WebSocket row split by transport, and the gRPC row still says protobuf is raw bytes (#739).
- `ClientConn`'s `h2_offer` now defaults to `Cleartext`; the honest end state is a required parameter with the three `server.cr` listener sites stamping it explicitly.

## Verification

`crystal tool format --check src spec` clean, `crystal build --no-codegen src/main.cr` clean, and the full suite on the merged result:

```
9343 examples, 0 failures, 0 errors, 0 pending
```

Every commit's specs were control-run — the change disabled, the new specs confirmed failing — because a green suite on its own proves nothing about whether a spec has teeth. #728 additionally has an end-to-end reproduction: 8.00 s client timeout before, 0.00 s after, against a Python origin answering from `handle_expect_100`.

---

한줄 요약: 프로토콜 지원 감사에서 살아남은 9건을 커밋 단위로 담았고, 가장 위험한 곳은 #728(요청 경로 3개의 흐름을 바꿈)·#733(새 서브시스템 + 스트림 수명주기 변경)·#731(보안 게이트 제거)입니다.


src/gori/proxy/conn/client_conn.cr
```diff
@@ -921,6 +1067,124 @@ module Gori::Proxy
...
+    #     conformant, and it is waiting for exactly the bytes the client is refusing to send, so
+    #     SOMEONE has to move first. If that write fails the client is gone: `{nil, false, true}`.
+    #     (Cost of the self-issued 100: a duplicate is possible when the origin's own 100 lands
+    #     just after the deadline. RFC 9110 §15.2 requires a client to tolerate 1xx it did not
+    #     even ask for, so a second one is harmless.)
+    private def settle_expectation(upstream : IO, req : Codec::RawRequest) : {Bytes?, Bool, Bool}
+      head = read_head_within(upstream, EXPECT_CONTINUE_WAIT)
+      if head
+        resp = Codec::Http1.parse_response_head(head)
+        return {head, false, false} unless interim_response?(resp) && !interim_has_body?(resp)
```
hahwul: **[bug]** `settle_expectation` treats any well-formed 1xx as settling `Expect: 100-continue` and immediately pumps the client body. 103 Early Hints (and 102 Processing) are interim but do **not** grant permission to send the withheld body — only 100 Continue or a final status does. After relaying a 103, gori blocks in `Codec::Body.stream` on a client that is still waiting for 100, while the origin’s subsequent 100/200 sit unread on the upstream socket. That is the same three-way stall #728 claims to close, just one 1xx later. A client with its own expect-timeout will eventually send; a client that waits for 100 will hang until some other timeout.

**Suggestion:** Loop here: relay every well-formed non-100 1xx, and only return `{nil, true, false}` on 100 (or after `EXPECT_CONTINUE_WAIT` with a self-issued 100). A final status should still take the `{head, false, false}` branch. Add a spec where the origin writes 103 then 100 while the client withholds the body.

src/gori/proxy/conn/client_conn.cr
```diff
@@ -921,6 +1067,124 @@ module Gori::Proxy
...
+    #     (The MAX_INTERIM run therefore restarts its count once, at one extra 1xx — a cap, not
+    #     a correctness boundary.)
+    #   - the origin answered something FINAL instead — 417 Expectation Failed, or a 401/403/30x
+    #     it can decide without reading the body, both of which RFC 9110 §10.1.1 explicitly
+    #     allows — then it does NOT want the body, and pumping one at a server that has stopped
+    #     reading is how a request gets smuggled into the next response's framing. Hand the head
+    #     back for relay and send nothing: `{head, false, false}`. A malformed 1xx (one declaring
+    #     a body) takes this branch too, on purpose: `skip_interim_responses` already knows how
+    #     to refuse it, and it gets a flow_id to record against, which this point does not have.
+    #   - the origin closed / errored before answering → `{nil, false, false}`. Uploading a body
```
hahwul: **[suggestion]** The comment documents a fourth outcome — origin closed/errored → `{nil, false, false}`, do not upload into a dead socket — but `read_head_within` collapses timeout, EOF, and reset into a single `nil`, and the `nil` path always self-issues 100 and pumps (`write_own_continue` at 1127). A reset/EOF origin therefore gets a 100 sent to the client and a body write into a dead upstream; `send_guard` then reports a generic acquire failure (and, if the 100 already went out, a 502 after a 1xx). The comment and the code describe different programs.

**Suggestion:** Either distinguish EOF from timeout in `read_head_within` and implement the documented `{nil, false, false}` path, or delete that bullet so the comment matches the timeout-only behaviour.

src/gori/tui/repeater_view/pointer.cr
```diff
@@ -108,6 +112,16 @@ class Gori::Tui::RepeaterView
...
   end
 
+  # GRPC RESPONSE: the one transcript pane that carries a chip — ` p:tree ` / ` p:bytes `,
+  # which picks the schema-less protobuf tree over a hex preview. `grpc_chip_x` and the label
+  # are the SAME derivation `render_grpc_chrome` draws from, so the live cells are exactly the
+  # painted ones. `limit:` is that draw's own stop.
+  private def grpc_chrome_hit(right : Rect, mx : Int32, my : Int32) : Symbol?
+    return nil unless @grpc_mode && right.w >= 2 && my == right.y
+    Frame.left_chip_hit(mx, my, right.y, grpc_chip_x(right),
+      [{:pretty, @pretty ? " p:bytes " : " p:tree "}] of {Symbol, String}, limit: right.right - 1)
```
hahwul: **[suggestion]** `grpc_chrome_hit` claims `limit:` is “that draw’s own stop,” but it passes `right.right - 1`. `render_grpc_chrome` tightens the stop to the latency meta’s left edge whenever `@result.duration_us` is set (`render_response.cr:63-67`) and skips the chip if it would overlap. On a half-width pane the hit test can therefore light `:pretty` on cells that show the duration, not ` p:bytes ` / ` p:tree ` — the same drawn-vs-hit split the comment says this helper exists to prevent.

**Suggestion:** Share one `stop` (or a small `grpc_chrome_limit(rect)`) between `render_grpc_chrome` and `grpc_chrome_hit`.

src/gori/interceptor.cr
```diff
@@ -341,13 +341,18 @@ module Gori
...
     # intercept is on.
 
-    # Is the sandbox on? Two callers, and neither is the Tunnel any more: it used to force every
-    # sandboxed connection to h1 so that `sandbox_blocks?` below would run per request, and #492
-    # step 4 replaced that with a per-stream refusal in `H2::StreamGate` (which is the other
-    # caller — it refuses a header block it cannot decode, since an unreadable head has no URL
-    # to test). The first is the CONNECT h2c refusal (`client_conn.cr`), where the raw h2c relay
-    # is wired with no gates at all, so there is nothing to gate per request and the whole tunnel
-    # is refused instead.
+    # Is the sandbox on? Every caller is now inside `H2::StreamGate`, and the Tunnel is not one:
```
hahwul: **[suggestion]** The `sandbox_enabled?` comment is a changelog of #492 / #549 / #731 (who used to call this, why the blanket CONNECT refusal existed, which PR removed it). That history is not a reason the method exists today and will be wrong the next time a caller is added.

**Suggestion:** Keep one sentence: StreamGate asks this when there is no URL to test, and it is independent of intercept-on. Move the rest to the PR or DESIGN.md §7.

src/gori/proxy/conn/client_conn.cr
```diff
@@ -34,14 +34,23 @@ module Gori::Proxy
   # read a line that did not exist. A refusal that names the wrong reason costs more time than
   # one that names none.
   enum H2Offer
-    # No tunnel made this decision (the plaintext listener, a blind CONNECT). Says nothing
-    # about a cause, on purpose.
+    # Nothing observed a cause. Says nothing about one, on purpose — the honest answer for a
+    # caller that has no observation to stamp.
+    #
+    # It has no producer today, and that is the point of #731 rather than dead weight: the only
```
hahwul: **[suggestion]** `H2Offer::Unknown` is now documented as having no producer, plus a paragraph on why the dead variant is kept for a hypothetical future caller. That is design-history, not a constraint on the next edit.

**Suggestion:** One line that Unknown is the causeless fallback. If it truly has no producer, drop the variant in a follow-up rather than narrating its retirement in-place.

src/gori/proxy/conn/client_conn.cr
```diff
@@ -464,8 +530,44 @@ module Gori::Proxy
         @sink.on_response(FlowMapper.error_response(flow_id, "client truncated request body"))
         return false
       end
-      handle_response(upstream, req, flow_id, started, host, port, scheme,
-        reused: reused, sent_head: sent_head, can_retry: retryable, sent_req: sent_req)
+      keep = handle_response(upstream, req, flow_id, started, host, port, scheme,
+        reused: reused, sent_head: sent_head, can_retry: retryable, sent_req: sent_req,
+        pre_read_head: early_head)
+      # The origin answered the expectation with a FINAL response (417, 401, a redirect…) and
```
hahwul: **[suggestion]** The post-`handle_response` `early_head` block spends ~30 lines explaining that the two closes are redundant today, not redundant as an invariant, and safe on the 101 branch that detaches `@upstream`. That is a design essay about a change that might happen, not a why the next reader needs at the call site.

**Suggestion:** Two lines: origin answered Expect with a final status and no body was sent, so neither leg is reusable; `release_upstream` is idempotent if the 101 path already detached the slot.

src/gori/proxy/h2/head_rewrite.cr
```diff
@@ -537,13 +548,51 @@ module Gori::Proxy::H2
       end
     end
 
-    # Which of the per-CONNECT gates would have wanted this stream — i.e. which host-scoped
-    # seams the h2 relay cannot reach have a live rule for the coalesced authority. Exactly the
-    # set `h2_candidate?` tests, asked about the authority instead of the CONNECT host, so the
-    # two cannot drift apart in what they consider unreachable.
+    # #730: the same three seams, on the ORDINARY stream — the one whose authority is the
```
hahwul: **[suggestion]** `notice_live_rule` opens with the #730 options memo (live-mutated tables, option (b) out of scope, lock-cost rationale, why a negative cannot be cached). Useful once in the issue; as a method comment it restates the code and the log line beneath it.

**Suggestion:** Cut to: same three unreachable kinds as `h2_candidate?`, spoken once per connection when a matching rule appears after CONNECT; do not cache a negative.

hahwul:
Review addressed. All three technical findings were real; I confirmed each against the code before changing anything.

## The bug — 103 Early Hints

Correct, and worse than "one 1xx later" in one respect: an origin that drips a 103 every `EXPECT_CONTINUE_WAIT - 1ms` would never technically time out either, so a per-read timeout would not have saved it.

`settle_expectation` now loops, relaying every well-formed non-100 1xx verbatim and reading on. Only `status == 100` concludes with "send the body"; a final status still takes the `{head, false, false}` branch; a malformed 1xx still falls through for `skip_interim_responses` to refuse with a `flow_id`.

Bounded twice over:
- **run length** — reuses `MAX_INTERIM` rather than inventing a second ceiling. Past it the body is withheld and the socket handed on as-is, so the remaining 1xx are still there for `skip_interim_responses` to refuse and record.
- **total time** — one `deadline` computed before the loop; each pass arms `read_head_within` with `deadline - Time.instant` only. Because the clock is read fresh, time spent *reading* a head is charged to the same budget.

Control run, per your suggested spec (origin writes 103 then 100 while the client withholds): with the 100-only gate reverted it fails `to include: "HTTP/1.1 100 Continue"`. A second control keeps the loop but re-arms a fresh wait per read — 9.75 s vs 1 s, the client stalled for the whole drip.

## `read_head_within` collapsing timeout, EOF and reset

Correct, and the comment did describe a different program. Took your option (a): the three are cheaply distinguishable — `read_byte` returning `nil` is a clean EOF, `IO::TimeoutError` is the timeout, any other raise is a reset. It now returns `Bytes | NoAnswer` (`Silent` / `Gone`), and `Gone` also covers a head that started and stopped. The documented `{nil, false, false}` is now produced, so no 100 is invented and no body goes into a dead socket.

One consequence worth naming: the post-`handle_response` guard keyed on `early_head`, which no longer covers every no-body-sent case. It keys on `unless send_body` now — exactly {final status, cap exceeded, dead origin}.

## `grpc_chrome_hit`'s `limit:`

Correct, and the comment asserted the very thing it did not do. Both sites now ask one `grpc_chrome_limit(rect)`; the label was also a literal copied into both files despite the same comment claiming it was shared, so that became `grpc_chip_label`. The "chip did not fit, so nothing was painted" case needs no extra branch — `Frame.left_chip_hit` breaks at the same chip the draw refuses.

The live window is 56–66 columns. Control run: both new specs fail with `Expected: :pretty to be nil` — the ghost toggle.

## The comment findings

Taken, with one deliberate exception. The rule I applied was *why the code is the way it is* (keep) versus *what changed and what used to be true* (cut), since dense explanatory comments with issue references are the house style here rather than an accident.

- `sandbox_enabled?` — the #492/#549/#731 changelog is gone; one sentence on the no-URL-to-test question remains, plus a pointer to `sandbox_blocks?` for callers that do have a URL.
- `H2Offer::Unknown` — 9 lines to 2. The variant is kept; retiring it is a separate change, not a narration.
- the post-`handle_response` block — ~30 lines to 6.
- `notice_live_rule` — the #730 options memo is gone. **I kept the lock-cost paragraph**, trimmed and re-headed `DO NOT LATCH THE NEGATIVE`. It is not history: it is the constraint that stops the next reader from caching a negative result, which would silently re-open the window the change closes. I had that paragraph subtly wrong once already and fixed it during review, which is the argument for it earning its place.

## Also, while here

- An HTTP/1.0 client sending `Expect: 100-continue` was entering the settlement and paying a full `EXPECT_CONTINUE_WAIT` against a quiet origin — a latency regression this PR introduced, for a client that was never waiting. The buffering path gated on the version and the streaming path did not. The test now lives in `expect_continue?` itself (RFC 9110 §10.1.1: the expectation MUST be ignored on a 1.0 request, so such a request is not withholding anything, which is what the predicate asks). Both now-unreachable guards inside `settle_expectation` are gone, along with its `req` parameter.
- CI caught a platform-dependent spec of ours: the unpinned h2c dial control raced gori's close, which is an RST on Linux and so an EPIPE on the client's own write, while macOS buffers it. Narrowed to `IO::Error` at that one call site; the pinned example is untouched.

Full suite on the merged result: `9349 examples, 0 failures, 0 errors, 0 pending`. Every new spec was control-run — the change disabled, the spec confirmed failing — since a green suite proves nothing about whether a spec has teeth.

---

한줄 요약: 기술 지적 3건 모두 실재해 수정했고(103 루프는 실행 예산까지 공유하도록), 주석 지적은 "왜 이런가"와 "무엇이 바뀌었나"를 기준으로 나눠 대부분 수용하되 `notice_live_rule`의 음성-latch 금지 문단만 남겼습니다.

Issue #743, opened 2026-08-17
fix(proto): a WebSocket over HTTP/2 still classifies as HTTP — PROTO shows HTTPS and `proto:ws` misses it


hahwul:
Found while fixing #742.

#733 captures the transcript and #742 made every reader show it, but a WebSocket opened over HTTP/2 is still classified as plain HTTP:

```crystal
# src/gori/proto.cr
def self.classify(status : Int32?, content_type : String?, request_content_type : String?) : Kind
  return Kind::Ws if status == 101
  ...
```

An RFC 8441 socket is `CONNECT` answered `200`, so it never takes that branch. Consequences:

- the History **PROTO** column shows `HTTPS`, not `WSS`, for a flow that has a full WebSocket transcript behind it;
- `proto:ws` in QL does not return it, so the filter an operator reaches for to find sockets silently omits every h2 one.

`Proto` documents itself as "the single source of truth the History PROTO column and the QL `proto:` field both defer to, so the label you see and the value you filter on can never drift". That contract still holds — both surfaces agree — they now agree on the wrong answer for one transport.

## Why this is not a one-line extension

`classify` takes `(status, content_type, request_content_type)`. Nothing in that triple distinguishes an extended CONNECT from an ordinary `CONNECT` to a proxy or from any other 200. The fact that makes it a WebSocket is the `:protocol: websocket` pseudo-header, which lives in the request head.

And the QL half cannot simply be widened: `QL.proto_cond` compiles to SQL over the flow table. It cannot read a request head, so whatever `classify` learns has to be reachable from a column.

So this needs a decision, roughly:

1. **Store it.** A column (or a reuse of an existing one) recording the negotiated application protocol at capture time, written by `H2::Assembler` where it already recognises `:protocol`, and by the h1 path at the 101. Then `classify` and `proto_cond` both read one column and the drift is closed structurally. Costs a migration.
2. **Derive it from the transcript.** "Has `ws_messages` rows" is what #742 settled on for the readers, and it is already a SQL-reachable fact. It answers slightly different question though — a socket that opened and carried nothing would classify as HTTP — and it puts a join in the History query's hot path.
3. **Read the request head in `classify` only**, and accept that `proto:ws` stays h1-only. This is the cheapest and it re-opens exactly the label-vs-filter drift the module exists to prevent, so it is listed to be rejected rather than chosen.

Option 1 looks right, but the migration and the h1/h2 write sites are worth agreeing on before anyone starts.

## Scope note

[...]

Issue #742, opened 2026-08-17
fix: the WS-over-HTTP/2 transcript is captured but no surface reads it — eight readers gate on `status == 101`


hahwul:
Follow-up to #733. The WS-over-h2 transcript is captured and persisted, but no surface will show it.

A WebSocket opened over HTTP/1.1 is a `101`. A WebSocket opened over HTTP/2 (RFC 8441 extended CONNECT) is a `CONNECT` answered `200`. Every reader asks `status == 101`, so an h2 socket's `ws_messages` rows are written to the store and then never read.

## The sites

| file:line | what it gates |
|---|---|
| `src/gori/tui/history_view.cr:1025` | History detail MESSAGES pane |
| `src/gori/tui/controllers/repeater_controller.cr:1394` | TUI "send to Repeater" seed |
| `src/gori/mcp/tools/flows.cr:69` | MCP `get_flow` |
| `src/gori/mcp/tools/repeater.cr:172` | MCP repeater seed |
| `src/gori/cli/run/history.cr:245` | `gori run history show` |
| `src/gori/cli/run/repeater.cr:294` | `gori run repeater` seed |
| `src/gori/cli/run/repeater.cr:1238` | seed, paired with `WsEngine.upgrade_request?` |
| `src/gori/export/har.cr:167` | HAR `websocket?` — its own comment says it asks this way *because* every other surface does |

`src/gori/export/har.cr:167`'s comment is the tell: the predicate was copied across surfaces because they all agreed, so they all drifted together the moment a second transport appeared.

## The honest predicate

"Does this flow have `ws_messages` rows", not "was it a 101". `Store#count_ws_messages` already exists (`src/gori/store/h2_frames.cr:129`).

Note this keeps #736 working: a non-WebSocket 101 whose only row is the `[gori]` opaque-upgrade notice still has rows, so it still renders.

## Not a mechanical find-and-replace

[...]

Issue #740, opened 2026-08-17
fix(proxy): `content_encoded?` gates only Content-Encoding, so a `Transfer-Encoding: gzip` body reaches the rule engine compressed


hahwul:
Found while fixing #735.

`ContentDecode.content_encoded?` is the wire gate that stops `ClientConn#apply_body_rewrite` from running Match & Replace body rules against compressed bytes. Its own docstring names the stake (`src/gori/proxy/codec/content_decode.cr:117-119`):

> which would hand a compressed body straight to the rule engine, the one outcome this predicate exists to prevent

It checks **only** Content-Encoding. `src/gori/proxy/codec/content_decode.cr:123-124`:

```crystal
def self.content_encoded?(head : Bytes) : Bool
  return false unless AsciiBytes.contains_ci?(head, CE_NEEDLE)
```

A `Transfer-Encoding: gzip` response carries no Content-Encoding at all, so it returns `false` on the very first line and the gate opens.

## What then happens

`src/gori/proxy/conn/client_conn.cr:1938-1946`:

```crystal
return {head, wire_body} if Codec::ContentDecode.content_encoded?(head)   # :1941 — passes
entity = framing.chunked? ? Codec::ContentDecode.dechunk(wire_body) : wire_body
rewritten = yield entity                                                  # gzip bytes handed to the rule engine
...
{reframe_to_length(head, rewritten.size), rewritten}
```

`Transfer-Encoding: gzip` (no `chunked`) is close-delimited by `Body.response_framing`, so `framing.chunked?` is false and `entity` is the raw DEFLATE stream. Two outcomes, both bad:

- **Usual case:** the rule pattern does not match compressed bytes, so it silently never fires. The operator sees a body rule that "doesn't work" on that host with nothing saying why.
- **Worse case:** the pattern *does* match — a short or binary-ish needle against a compressed stream is a byte-value coincidence, which is exactly the risk the docstring at `:110-113` describes. The rewritten gzip stream is then reframed by `reframe_to_length`, which drops Transfer-Encoding, so the client receives raw DEFLATE bytes advertised as an identity `Content-Length` body. Corrupted response, no error.

The asymmetry is stark: the predicate goes to real lengths for Content-Encoding — it fails **closed** on an obs-folded `Content-Encoding:\r\n gzip` (`:114-122`, `:130`) precisely so a hidden coding cannot slip through — while the sibling header family has no gate at all.

## Fix

[...]

Issue #733, opened 2026-08-17
feat(proxy): WebSocket over HTTP/2 (RFC 8441 extended CONNECT) is relayed but never captured


hahwul:
A WebSocket opened over HTTP/2 — RFC 8441 extended CONNECT, which is what modern browsers use when the origin allows it — is relayed but never read. No `ws_messages` rows, no message gate, no Match & Replace, no Repeater seed.

## What exists today

`src/gori/proxy/h2/assembler.cr:525-581` recognises the `:protocol` pseudo-header and writes an honest per-flow advisory (`:551-556`, `:576-581`). That part is good engineering — gori knows what it is looking at and says so. But the capability behind it is absent:

- the socket lands as one opaque `DATA` blob, capped at 2 MiB;
- none of the WebSocket machinery that `src/gori/proxy/ws/` provides on the h1 upgrade path is reachable.

## gori actively enables the path it cannot read

`src/gori/proxy/h2/assembler.cr:528-531` documents that the origin's `SETTINGS_ENABLE_CONNECT_PROTOCOL` is relayed verbatim. So a browser facing an 8441-capable origin is *entitled* to take the h2 path — gori tells it that it may, then cannot follow it there.

## It is nowhere in the docs

`grep -rni "8441|extended connect|:protocol"` over `README.md`, `docs/` and `CHANGELOG.md` → **zero hits**. `docs/content/guide/proxy.md`'s protocol table lists WebSocket as "Live message capture, repeater, per-message intercept … and Match & Replace on messages" with no h1/h2 qualifier. The per-flow advisory is the only place an operator learns otherwise, and only after the fact, one flow at a time.

## Scope

Two options, and the docs half is worth doing either way:

- **Cheap:** an option to strip `SETTINGS_ENABLE_CONNECT_PROTOCOL` from the origin's SETTINGS so the client falls back to the h1 `Upgrade:` path that gori *can* read in full. Note this requires parsing SETTINGS, which the relay deliberately never does today (`src/gori/proxy/h2/head_rewrite.cr:57-61`, `src/gori/proxy/h2/stream_gate/body.cr:130-132`) — so it is a real design decision, not a one-liner.
- **Full:** run the WS frame codec over the extended-CONNECT stream's DATA frames so h2 sockets get the same transcript, gate and rules as h1 ones.
- **Docs, regardless:** qualify the WebSocket row in the protocol table, and say what the advisory means.

---

[...]

PR #744, opened 2026-08-17
Protocol gaps, wave 3: the WS-over-h2 follow-through, the TE rewrite gate, and the docs

### LINKED ISSUE #739: docs: the proxy guide's protobuf and WebSocket-compression claims are both stale

hahwul:
The follow-ups #741 left behind. Closes #739, #740, #742.

#741 landed nine protocol fixes and produced two of these itself: WS-over-HTTP/2 capture that no surface could read, and a `Transfer-Encoding` decode path whose sibling wire gate had not moved with it.

## `a0e21b45` — #742: a WebSocket is read from its transcript, not from a 101

#733 captured RFC 8441 sockets and wrote `ws_messages` rows. But an extended CONNECT is answered `200`, and eight readers asked `status == 101`, so the transcript was stored and shown by nobody. `export/har.cr`'s own comment named the mechanism: it asked that way *because every other surface did*, so all eight drifted together the day a second transport appeared.

**Display and export now hold no predicate at all.** `Store#ws_messages(id)` already returns empty for a non-socket, so the honest question — are there rows? — is answered by the rows. One structural predicate survives, `Store::FlowDetail#websocket?`, used only where the rows *cannot* answer: telling "a socket with no transcript" apart from "an ordinary HTTP flow" in `Har.skip_reason`, and the repeater refusal.

**The three repeater seeds refuse, with a stated reason.** Not because the frames differ — they are the same RFC 6455 frames — but because the handshake does. `WsEngine` writes an h1 `Upgrade:` and accepts nothing but a 101; an h2 capture holds `CONNECT /chat HTTP/2` and no `Upgrade:` header. Seeding it would hand over a send that cannot connect, and making it connect would mean gori inventing a `Sec-WebSocket-Key` the client never sent and storing it under that flow's provenance. The TUI opens the ordinary HTTP tab instead (the CONNECT *is* a real h2 request) and its status line points at History's MESSAGES pane.

**HAR** writes the real `CONNECT`/`200` with `_webSocketMessages` beside it. Synthesizing a `101` so the entry looked like the h1 one would put a response on the record no origin sent — and `Import::Har` would read it straight back as a flow that never existed.

One behaviour change beyond h2: a non-WebSocket 101 (#736's kubectl SPDY) no longer seeds a WebSocket repeater tab. It used to seed an empty one whose send would have written `Upgrade: SPDY/3.1` plus a fabricated key and then read SPDY bytes as RFC 6455 frames.

## `f36591c6` — #740: the body-rewrite gate consults both header families

`ContentDecode.content_encoded?` stops Match & Replace body rules from running against compressed bytes — its docstring calls that "the one outcome this predicate exists to prevent" — and it checked only `Content-Encoding`. A `Transfer-Encoding: gzip, chunked` response carries no `Content-Encoding`, so it returned false on the first line and the gate opened. The control run prints the consequence directly: with the fix reverted, the client receives `Content-Length: 9\r\n\r\nCLOBBERED`.

Two things worth review attention:

- **The reachable shape is `gzip, chunked`, not bare `gzip`.** A close-delimited `Transfer-Encoding: gzip` never reaches `apply_body_rewrite` at all — `buffer_response_body?` requires length or chunked framing. The predicate refuses both regardless; the issue as filed overstated the reach.
- **The fail-closed rule had to be the broad form.** `Transfer-Encoding: chunked,\r\n gzip` parses as the value `"chunked,"`, whose token list is exactly `["chunked"]` — indistinguishable from ordinary chunked framing, while a lenient backend reads the gzip. Since `encoding_headers` skips colon-less lines, nothing can attribute a continuation line to its header, so a narrow test is not available. The cost is a rewrite skipped on an already-malformed head, with the response staying byte-exact.

A plain `Transfer-Encoding: chunked` still allows the rewrite — that regression guard passes in both the fixed and the control run, which is what it is for.

## `2fb74d78` — #739: four stale claims in the proxy guide

The gRPC row still said protobuf was "raw bytes (no `.proto` schema)" two releases after the schema-less decoder shipped, and nothing in `docs/` mentioned the decoder at all. The WebSocket compression paragraph offered TLS passthrough as the only way to keep a socket byte-exact, when the strip is ordered before Match & Replace *precisely* so a rule can restore the offer — which, unlike passthrough, keeps the flow captured. Plus the WebSocket row now splits by transport, and the compressed-body caveat covers both header families.

EN/KO parity kept (74 pages, 37:37); `hwaro build` renders both new anchors in each language.

## Known caveat, filed as #743

`Proto.classify` still keys on `status == 101`, so a WebSocket over HTTP/2 shows `HTTPS` in the PROTO column and `proto:ws` does not match it — with its transcript visible in the MESSAGES pane. It is not a one-line fix: the fact that makes it a WebSocket is the `:protocol` pseudo-header in the request head, and `QL.proto_cond` compiles to SQL and cannot read a head. The docs state the caveat and the workaround rather than leaving an operator to discover it.

## Verification

`crystal tool format --check src spec` clean, `crystal build --no-codegen src/main.cr` clean, full suite on the merged result:

```
9376 examples, 0 failures, 0 errors, 0 pending
```

Every new spec was control-run — the change reverted, the spec confirmed failing. For #742 that is 9 of 20 examples going red, with the 11 h1 / #736 / negative guards staying green either way.

Note on verifying this branch locally: use `crystal build --no-codegen src/main.cr`, not `src/gori.cr`. The library root has no top-level execution, so Crystal does not typecheck a method body nothing calls — it accepted a non-compiling merge during #741.

---

한줄 요약: #741이 남긴 후속 3건으로, WS-over-h2 전사를 실제로 읽게 하고, 압축 본문 rewrite 게이트를 Transfer-Encoding까지 확장하고, 프로토콜 문서의 낡은 주장 네 개를 코드에 맞춰 고쳤습니다.


hahwul:
Three more commits on this branch since it was opened. Also closes #743 and #745.

## `6fe9ee82` — #743: a WebSocket over HTTP/2 is a WebSocket in the PROTO column too

#742 made every reader show the h2 transcript, but `Proto.classify` still keyed on `status == 101`, so the flow read `HTTPS` and `proto:ws` did not return it. The `#739` docs commit above had to write that caveat down; this removes it, and the paragraph is rewritten in both languages.

**The design was not open** — I framed it in the issue as three options, and then found V14 had already answered it. `store/schema.cr:946-968` fixed the identical shape for gRPC (a call classified as gRPC exactly when it *succeeded*) and states the constraint verbatim: a column, not a read-time re-parse, because `QL.proto_cond` compiles `proto:` to SQL and a label derived from bytes the query cannot see is the drift `Proto` exists to prevent. V16 follows it, NULL semantics included — pre-migration rows stay unclassified rather than being guessed at.

`flows.connect_protocol` holds the extended CONNECT's `:protocol` token verbatim, so `connect-udp` (RFC 9298) and `connect-ip` (RFC 9484) stay distinguishable from `websocket` on disk rather than being flattened into a boolean. It is written from the h2 decoder's own read of the pseudo-header, **not** lifted off the stored head — a pseudo-header reaches the head only as a synthetic `X-Gori-Protocol` marker, which an imported HAR can forge.

Two deliberate asymmetries, both argued in the commit: a *refused* h2 handshake classifies as HTTP, matching a refused h1 one and matching `FlowDetail#websocket?`; and `classify` takes the new field as a **required** parameter, because a default would have left all four call sites compiling untouched and still printing `HTTPS`.

## `6f9e17df` — #745: a refused body rule says so

A Match & Replace body rule that is live, enabled and correctly scoped simply did not fire on a compressed body, and nothing anywhere said why. The refusal is right; the silence was not. #740 widened that branch from `Content-Encoding` to `Transfer-Encoding`, so more responses take it now.

It lands on `FlowRow#advisory` — whose own docstring already named this case as the reason the column exists — not the `[gori]` transcript seam, which is a row *positioned in a stream of frames* and has no meaning on an ordinary HTTP flow. So it rides History, `gori run show`, MCP `get_flow` and export for free.

Rated per {direction, host, coding} within the connection rather than per connection: a cleartext forward-proxy `ClientConn` serves whatever hosts the client's keep-alives name, so a per-connection latch would let the first host silence every other one.

## `cd1796aa` — `h2_offer` becomes required

Flagged during #731. It defaulted to `Cleartext`, correct by observation but silent if a future caller inherited it. Now required and stamped at every site — which turned up a **fourth** cleartext listener that #731's "three" predates: the transparent h2c entry added by #737 on this same branch.

## `d41e3daf` — a follow-up the #745 author flagged rather than forced

`flow_mapper.cr` was owned by the parallel #743 work, so #745 set its advisory by rebuilding each DTO field-by-field in two private overloads. Both changes have landed, so `FlowMapper.request`/`.response` now take `advisory:` and the 27 lines of hand-copied fields are gone. Their only failure mode was silence: a field added to either DTO and not to the copy would have been dropped whenever an advisory was set.

## Verification

`9393 examples, 0 failures, 0 errors, 0 pending`; format and `crystal build --no-codegen src/main.cr` clean.

Every new spec control-run. #743's second control drops V16 from `MIGRATIONS` alone and gets `no such column: connect_protocol`, so the migration is load-bearing rather than merely present. #745's second control removes only the log latch, to prove the rate claim is tested and not just asserted.

---

한줄 요약: #743(WS-over-h2를 PROTO/`proto:ws`에서도 WebSocket으로)과 #745(압축 때문에 거부된 body 규칙을 flow에 기록)를 더했고, `h2_offer` required 전환과 `FlowMapper` advisory 배선 정리까지 포함합니다.

Issue #749, opened 2026-08-18
feat: Repeater/Fuzz can record History, and fuzz can seed from a repeater id


hahwul:
A headless tester loop dies here:

```text
gori run repeater create --target http://127.0.0.1:18765 --request-raw '...'
gori run repeater send 1          # 200, body looks right
gori run history                  # no flows
gori run sitemap                  # no endpoints
gori run fuzz 1                   # no flow #1
gori run show 1                   # no flow #1
```

MCP `send_request` already defaults `record_history:true`. MCP `fuzz_start` has `record_history: none|matched|all`. CLI `repeater send` and `gori run fuzz` have neither, and fuzz only seeds from `--flow` or `--request FILE` — not from a repeater session.

TUI can `Space → f` (repeater.fuzz). Headless has to rewrite the request to a file. `gori run compare` only takes flow ids, so a Repeater send cannot enter a CLI diff without a History row either. The playbook is honest about this ("neither leaves a captured flow behind") — that is the gap, not a surprise.

## Why it is worth doing

The workbench (Repeater/Fuzz) and the evidence store (History/Sitemap/HAR/compare) are two universes. That is fine as a *default* — Burp Repeater also stays out of Proxy history — but the operator needs an explicit way to punch a hole when they want a flow id for the next tool.

## Proposal

1. **`gori run repeater send --record-history`** (and the TUI equivalent, off by default) writes the outbound request + response as a History flow and prints the new `flow_id`. MCP already has the write path.
2. **`gori run fuzz --record-history=none|matched|all`**, same enum as MCP `fuzz_start`. Default stays `none`.
3. **`gori run fuzz --repeater=ID`** (MCP `fuzz_start.repeater_id`) seeds the template from a saved HTTP repeater the way `--flow` seeds from a capture. WS repeaters are out of scope here — refuse with a named reason.

Do not silently start recording every Repeater/Fuzz send. The default stays workbench-only; the flag is the punch-through.

---

한줄 요약: Repeater/Fuzz 전송을 History에 *옵트인*으로 남길 수 있게 하고, `fuzz --repeater=ID`로 리피터 세션에서 바로 퍼즈하자는 이야기입니다.

hahwul:
Fixed in #753 (`4b55b462`), merged to `main`. All three proposals, with the default kept where you put it.

[...]

Issue #748, opened 2026-08-18
feat(cli): opt-in in-scope filter for History/Probe — capture stays everything


hahwul:
Capture-everything is the right default. A busy browser session still makes the *view* unusable: `gori run history` opens on Chrome/Google noise, and `gori run probe --severity high` ranks `accounts.google.com` CORS above the target. The TUI already has the answer — `s` is a *lens*, not a gate — but the CLI/QL side has no equivalent.

This is **not** a request to stop recording out-of-scope traffic, to turn the lens on by default, or to make Probe skip out-of-scope hosts unless asked. Everything is captured. The operator turns a filter on when the list is too loud.

## What already exists

- TUI: `s` / the `scope:N` chip toggles the History/Sitemap/Probe *lens*. Off by default, even after include rules are added (`Scope filtering: DISABLED`).
- `gori run sitemap --in-scope` already filters whole hosts via `Scope#host_in_scope?`.
- QL has no `scope:` field. MCP `list_history` / `probe_scan` have no `--in-scope`.

## Proposal

1. **`gori run history --in-scope`** and **`gori run probe --in-scope`**, same contract as sitemap: opt-in, host-level, no-op (with a stderr note) when no include rules exist.
2. **QL `scope:in` / `scope:out`** (or `in-scope:true`) so the same predicate works in the TUI filter bar, colour rules, and MCP `query`. Keep it a filter term, never a default.
3. Do **not** change the capture path, the sandbox, or the active-tool scope gate. Those are different questions.

`--in-scope` without any include rule should say so and return empty, matching sitemap today — not silently show everything.

---

한줄 요약: 캡처는 그대로 전부 남고, History/Probe/QL에서만 TUI `s` 렌즈처럼 스코프를 *켜서* 걸 수 있게 하자는 이야기입니다. 기본값을 바꾸지 말 것.

hahwul:
The surface half is fixed in #753 (`e842d14a`), merged to `main`. The QL half is split out as #754 — see below.

**Landed, exactly as filed — opt-in, and capture untouched:**

- `gori run history --in-scope` composes the same per-flow include/exclude filter the TUI History lens applies.
- `gori run probe --in-scope` filters the report by `host_in_scope?`, matching the TUI Probe lens. Every flow is still scanned; it narrows the report, and it is independent of `--active`/`--allow-unscoped`, which gate what gets SENT.
- MCP `list_history` and `probe_scan` gain `in_scope`.
[...]

Issue #747, opened 2026-08-18
feat(jwt): CLI claim edit (--payload / --set) and show the JWT tab by default


hahwul:
The JWT *TUI* loop is complete: send a token to the tab, edit PAYLOAD JSON on the Encode lens, re-sign, fire `--attacks`. Two things keep a tester from using that loop.

## 1. `gori run jwt --encode` cannot change claims

```bash
gori run jwt "$TOKEN" --encode --secret secret
```

re-signs the *same* payload. There is no `--payload`, `--set role=admin`, or `--claim`. The playbook's headless example only re-signs; the "escalate a `role`" step is TUI-only. MCP `jwt_encode` should grow the same knobs if it does not already have them.

## 2. The JWT tab is hidden by default

The playbook says to reveal it from the tab-bar `⋯` menu, the palette (**Go to JWT**), or Preferences. A tester who has just captured `Authorization: Bearer eyJ…` will not look in `⋯`.

## Proposal

1. **CLI / MCP encode accepts an edited payload:**
   - `--payload '{...}'` replaces the claims object wholesale
   - `--set role=admin` (repeatable) patches one claim on the decoded payload, then signs
   Keep `--alg` / `--secret` as they are. Refuse `--payload` and `--set` together.
2. **Show the JWT tab in the default tab set** (same class as Decoder / OAST — visible, not tucked behind `⋯`). Operators who never touch JWTs can still hide it from Preferences; first-run should not hide the workbench the playbook depends on.

Out of scope: cracking HS secrets beyond the existing `--attacks` weak-secret list; editing the header JSON (kid/jku already live under `--attacks`).

---

한줄 요약: `gori run jwt --encode`가 클레임을 바꿈 수 있게 하고, JWT 탭을 기본으로 보이게 하자는 이야기입니다.

hahwul:
Fixed in #753 (`d22186ba`), merged to `main`.

Both halves:

[...]

PR #753, opened 2026-08-19
The four issues still live in #729-749: non-HTTP flows, an in-scope lens, workbench→History, and JWT claim edits

### LINKED ISSUE #729: fix(proxy): a non-HTTP protocol through the MITM dies silently — no flow, no log, and nothing names TLS passthrough
Point a client that speaks a non-HTTP protocol at gori and the connection dies with no History row, no `gori.log` line, no TUI chip, and no error message. Nothing anywhere names `network.tls_passthrough`, which is the remedy.

## What actually happens

Traced end to end and reproduced against a real `Gori::Proxy::Server`. `Tls::Tunnel#intercept` (`src/gori/proxy/tls/tunnel.cr:185-201`) hands the decrypted socket to `ClientConn#run`, which calls `Codec::Http1.read_head` (`src/gori/proxy/conn/client_conn.cr:246`). `read_head_deadlined` (`src/gori/proxy/codec/http1.cr:56-77`) scans byte-by-byte for `CRLFCRLF` — bytes that never contain it simply block.

| input | proxy answers | flows recorded |
|---|---|---|
| MQTT 3.1.1 CONNECT (`0x10 0x0c 4d 51 54 54 …`) | **nothing** — blocks to `HEAD_DEADLINE` = 30 s, then closes silently | **0 / 0** |
| server-speaks-first (SMTP, IMAP, MySQL — the client sends nothing) | **nothing** — blocks to `CLIENT_IO_TIMEOUT` = 30 s (`src/gori/proxy/socket_tuning.cr:34,40`); **the upstream is never dialled** | **0 / 0** |
| binary that happens to contain `CRLFCRLF` | `502 Bad Gateway` | 1 flow reading `"upstream connect failed: :80"` — an error naming a host that was never the target |

The `IO::TimeoutError` unwinds into `ClientConn#run`'s blanket rescue at `src/gori/proxy/conn/client_conn.cr:202-207`, which logs nothing and records nothing.

## Two more entry points with the same silence

**Plaintext SSH through CONNECT.** `handle_connect`'s first-byte peek (`src/gori/proxy/conn/client_conn.cr:1391-1416`) only separates `0x50` (the h2c preface) from *everything else, which it assumes is TLS*:

```crystal
if first == 0x50_u8 then intercept_h2c(...) else tls.intercept(host, port, stream, @sink, ...) end
```

So `ssh -o ProxyCommand='nc -X connect …'` — a routine corporate-proxy pattern — feeds `"SSH-2.0-OpenSSH…"` into an OpenSSL **server** handshake, which raises into `Tunnel#intercept`'s bare rescue (`src/gori/proxy/tls/tunnel.cr:203-205`) and closes instantly. No flow, no log. Worse: `reflect_origin_h2` (`src/gori/proxy/tls/tunnel.cr:237-259`) has already fired a bogus TLS ClientHello advertising ALPN `h2` at the SSH server's port 22 before failing.

**The transparent listener.** `src/gori/proxy/server.cr:279-294` routes on `0x16` only; cleartext non-HTTP is *not* dropped, it goes straight to `ClientConn` — i.e. the same 30-second silent hang. This matters more here than anywhere else, because a firewall redirect sweeps in all TCP by design, not just proxy-aware clients.

(The two nearby drops are different things and are both fine: `server.cr:358-366` is TLS with no SNI *and* no kernel `SO_ORIGINAL_DST`, announced once via `log_missing_sni` at `:411-415`; `server.cr:371` is the sandbox.)

## Why this is a defect and not a documented limitation

`#493` / `#494` were filed for exactly this shape — *"the h2→h1 ALPN downgrade is completely silent"* / *"a gRPC connection killed by the ALPN downgrade never names the gate that caused it"* — and produced the whole `H2Offer` + `refusal_reason` machinery (`src/gori/proxy/conn/client_conn.cr:36-74`, `:271-275`). That fixed it for the one input the parser can recognise (`PRI * HTTP/2.0`). Non-HTTP bytes are the same bug with a different input.

The remedy is also undiscoverable. `docs/content/guide/proxy.md:395-414` frames passthrough exclusively as the *"When a Pinned App Is in the Way"* answer. `grep -riE "mqtt|smtp|imap|non-http|raw tcp"` over `docs/`, `README.md` and `DESIGN.md` returns nothing on this subject, so an operator whose MQTT app hangs has no reason to look there.

## Why an auto-fallback is the wrong fix

Tempting, but no:

1. **It would not be passthrough.** TLS is already terminated under gori's leaf and the ClientHello has been consumed and answered. A "fallback tunnel" is gori decrypting one leg and re-encrypting the other — still a MITM, just uncaptured. A pinning client has already failed by that point, so it buys nothing for the case passthrough exists for.
2. **Silent tunneling is the anti-pattern this codebase names.** `src/gori/settings/network.cr:60-64`: *"A bypassed host is otherwise INVISIBLE … so 'why is this host missing from History?' has no answer anywhere."*
3. **"Not HTTP" cannot be separated from "slow HTTP" by timeout** without weakening the slowloris defense `HEAD_DEADLINE` exists for.

## Proposed scope

A cheap, timeout-free, false-positive-free detector decides on the first line:

- any byte in the request-line region that is not a `tchar`/SP/HTAB → definitively not HTTP (catches MQTT `0x10`, AMQP `\x00`, TLS-in-TLS `0x16`, binary RPC);
- the first line completes at CR/LF but does not end in `HTTP/x.y` → definitively not HTTP (catches SSH, SMTP, FTP).

Neither waits, and neither affects legitimately drip-fed HTTP. Then:

1. On detection, `record_error` a visible flow whose message names the observed shape **and names `network.tls_passthrough` as the remedy** — `H2Offer#refusal_reason` is the template.
2. Give the same treatment to the head-deadline / idle-timeout expiry in the blanket rescue at `client_conn.cr:202-207`. A server-speaks-first protocol is the one case that must still use the timeout.
3. `handle_connect`'s peek must not feed obviously-non-TLS first bytes into an OpenSSL server handshake, and `Tunnel#intercept`'s bare rescue should record why it closed.
4. Docs: name non-HTTP protocols as a second reason to reach for passthrough.

---

한줄 요약: MQTT·SSH·SMTP 같은 비-HTTP 프로토콜이 MITM을 지나면 30초 무음 행 후 flow도 로그도 없이 죽는데, 해결책인 `network.tls_passthrough`를 알려주는 곳이 어디에도 없습니다.

hahwul:
The four issues that survived a validity re-check of #729-749. Closes #729, #747, #748, #749.

Sixteen issues in that range were open; twelve were already fixed by #741/#744 and had simply never been closed (verified against the code, then closed with the commit that fixed each). These four were still real.

## The commits

| commit | issue | what |
|---|---|---|
| `d22186ba` | #747 | `gori run jwt --encode` can change claims (`--set`, `--payload`), MCP `jwt_encode` gains `set`, JWT tab is visible by default |
| `e842d14a` | #748 | `--in-scope` on `history`/`probe` and MCP `list_history`/`probe_scan` — capture still records everything |
| `4b55b462` | #749 | `repeater send --record-history`, `fuzz --record-history=none\|matched\|all`, `fuzz --repeater=ID` |
| `c9c3fa39` | #729 | a non-HTTP protocol on the proxy path is recorded instead of hanging 30 s in silence |
| `42a59189` | — | eight review findings on the above, one of them a P7 regression — see below |

## Where to look hardest

**`c9c3fa39` + `42a59189` (#729) — a hot path, and the one place this went wrong twice.**

The detector lives inside `read_head`'s deadline loop, and that loop reads **both** request and response heads (`safe_read_head`). Enabled unconditionally, it classified `HTTP/1.1 200 OK` as non-HTTP — the last token is `OK` — and **truncated every response head**. It surfaced only as `content_type` being nil in the SSE integration test; the codec unit tests were green throughout. It is now gated on `detect_non_http`, passed by the single client-request call site.

The second, found in review: the first draft required the first line's last token to be a literal `HTTP/<d>.<d>`. That refuses `HTTP/1.10`, a lowercase `http/1.1`, HTTP/0.9 and a bare `GET` — the deliberate parser-differential payloads this codebase exists to carry byte-exact (P7), and `parse_request_head` keeps every one. gori would have closed the connection and blamed a TLS-passthrough setting for the operator's own test.

So the detector is now **one signal**: a non-token first byte (MQTT `0x10`, AMQP `0x00`, a TLS ClientHello `0x16`, binary RPC). A **text** banner — SSH, SMTP — is an explicit non-goal: on the first line it cannot be told from a malformed request line, so those still wait out the head deadline exactly as before. That is stated in the code, pinned by a spec, and written into the config reference rather than left for someone to "improve" later. `spec/proxy/server_spec.cr` carries the complement: a `HTTP/1.10` request must still reach the origin byte-exact.

**`4b55b462` (#749) — two universes, deliberately.** A Repeater/Fuzz send leaves no History flow by default (Burp Repeater stays out of Proxy history). The flags are the punch-through, not a new default. Two shared engine recorders — `Repeater::HistoryRecord` (Plan+Result) and `Fuzz::HistoryRecord` (Result) — so CLI and MCP project the same flow; review finding 4 is why MCP now *delegates* to the latter instead of holding a byte-identical copy under a comment claiming they were shared.

**`e842d14a` (#748) — deliberately partial.** `Scope#filter(force:)` applies the include/exclude rules regardless of the persisted ⇧S flag, which is what an opt-in `--in-scope` needs. The QL `scope:in`/`scope:out` half of the issue is **not** here: threading scope context into the pure grammar touches two backends (`QL` + `InterceptFilter`) and colormarker's two eval tiers, and that file is built around not letting a field silently broaden a standing rule. It wants its own design pass. Filing a follow-up.

## Review findings, and what changed

All eight were real. Beyond #1 above:

- `fuzz --record-history` on a **project-less** run (`--request`/stdin, no `--project`) resolved the *ambient default project* and wrote up to 5 000 flows into it — two lines after printing "no project in play … sending UNSCOPED". Now refused up front.
- `repeater send --record-history --format json` printed a **second top-level JSON object**, breaking `| jq`, and a bare integer line in text mode. The id is now a `recorded_flow_id` field inside the one object; text mode says it on STDERR.
- A recording that failed **every** write reported `recorded 0 flows` with no reason. The rescue was bare — and the commonest failure (a rolled-back insert returning `0`) was not even an exception. Both paths now report through an `on_error` block, bounded per job on MCP, once per run on the CLI.
- MCP `fuzz_start` silently ignored `repeater_id` when `template`/`flow_id` was also given; it now refuses the pair, as the CLI already did.
- The `--repeater` seed dropped the session's stored **SNI**, so a pinned session swept a different vhost than `repeater send N` reaches. Threaded on both surfaces; an explicit `--sni` still wins.
- `--record-history` + `--format json` retains each row's bytes in the end-of-run buffer; the note points at `--format jsonl`, which streams.

## Deliberately not here

- **QL `scope:`** (#748's second half) — above.
- **#729's server-speaks-first case** (SMTP/IMAP/MySQL, where the client sends nothing): there are no bytes to classify, so it still waits out the idle timeout.
- **#729's CONNECT peek** feeding obviously-non-TLS first bytes into an OpenSSL *server* handshake, and `Tunnel#intercept`'s bare rescue. Both are the TLS path, not the head reader.

## Verification

`crystal tool format --check src spec` clean, `crystal build --no-codegen src/main.cr` clean (the root that actually typechecks — `src/gori.cr` accepted a non-compiling merge during #741).

```
1576 examples, 0 failures   (fuzz, repeater, mcp, cli, scope, jwt, proxy/codec)
  53 examples, 0 failures   (proxy/server — socket-bound, run alone)
 134 examples, 0 failures   (the remaining proxy specs)
```

Every behavioural change was also driven end-to-end against a local origin: the JWT `--set` round-trip, `history`/`probe --in-scope` against a two-host capture with the ⇧S lens off, `repeater send --record-history` → `history` → `fuzz --repeater`, and the MQTT/TLS non-HTTP flows.

One environment note for anyone re-running this locally: two runaway `crystal-run-spec` processes left by earlier sessions (85+ CPU-hours each, on the default `~/.cache/crystal` rather than a `CRYSTAL_CACHE_DIR`) made socket specs look like they were hanging on a logic bug. `ps aux | grep crystal-run-spec` and kill by absurd TIME before trusting a "hang".

---

한줄 요약: #729-749에서 아직 유효한 4건을 구현했고, 코드 리뷰 8건도 함께 고쳤습니다 — 그중 하나는 제가 넣은 P7 회귀(malformed 요청 라인을 비-HTTP로 거부)라 탐지기를 "첫 바이트가 이진" 하나로 좁혔습니다.


hahwul:
Follow-ups filed for the two things this PR deliberately leaves out, so neither is only recorded in a PR body:

- **#754** — QL `scope:in` / `scope:out`, the second half of #748. Needs scope context threaded into the pure grammar, plus decisions in `InterceptFilter` and colormarker's two eval tiers.
- **#755** — the two #729 cases on the TLS path: the silent server-speaks-first timeout (no bytes to classify), and `handle_connect`'s peek treating "not `0x50`" as TLS, which feeds an SSH banner into an OpenSSL *server* handshake after already firing an ALPN probe at port 22.

PR #759, opened 2026-08-20
fix(store): recover from a two-writer collision instead of wedging on it (#752)

### LINKED ISSUE #752: TUI + MCP using the same project can repeatedly hit "database is locked" and stop saving flows
## Summary

When the gori TUI and `gori mcp` use the same project at the same time, gori can start repeating `database is locked` warnings and stop saving newly captured flows. The TUI remains open, but persistence does not recover until the affected process is restarted.

This happened during normal TUI + MCP usage, including with MCP `--read-only`.
The project database remained readable and `PRAGMA quick_check` returned `ok`.

## Environment

- gori: v0.3.2
- current `main` checked at: `f3a5f9a9f252fdd20f13e980d5bdda4c337755a0`
- OS: macOS / Apple Silicon
- Crystal: 1.21.0
- crystal-db: 0.14.0
- crystal-sqlite3: 0.23.0
- SQLite: 3.51.0
- journal mode: WAL
- processes: one TUI and one MCP process accessing the same project database

MCP was started with `--read-only`. That option gates MCP action tools, but the Store still opens a writable SQLite connection and starts its background FTS writer.

## Analysis Note

I’m not deeply familiar with Crystal or its database libraries, so I used Codex to help inspect the relevant source code and build a local reproducer.

The behavior and SQLite result codes described below were observed locally, but my interpretation of the underlying cause may be incomplete or incorrect.

## Suspected Root Cause

`Store#index_pending_batch` currently reads dirty rows before beginning its deferred transaction. A peer can commit after that read and before the first FTS write.

The observed statement state is consistent with a wrapper cleanup path that could cause this otherwise recoverable race to persist until the process is restarted:

- crystal-sqlite3 0.23.0 `Statement#perform_exec` calls `sqlite3_reset()` before `sqlite3_step()`, but does not reset or finalize the statement after `sqlite3_step()` fails.
- crystal-db caches that prepared statement.
- Rolling back the transaction does not reset the failed FTS statement itself.
- The old read snapshot can therefore remain attached to the connection, causing a later write to receive `SQLITE_BUSY_SNAPSHOT`.

This would explain the observed combination of a ~250 ms FTS retry loop, failed normal Store batches, a healthy DB, and recovery after process restart.

hahwul:
Closes #752.

## What was happening

A TUI and a `gori mcp` on one project are two SQLite writers. A collision between them was **terminal rather than momentary**:

1. The losing write hits `SQLITE_BUSY` / `SQLITE_BUSY_SNAPSHOT`.
2. crystal-sqlite3's `Statement#perform_exec` resets *before* `sqlite3_step` and never after a failed one, so the statement is left in progress.
3. SQLite then refuses the ROLLBACK.
4. crystal-db's `TopLevelTransaction` clears its in-transaction flag only once ROLLBACK *returns*, so the connection stays marked as in a transaction.
5. The writer pinned **one** connection with `using_connection` for the life of the store — so every later write failed the same way, and the open transaction kept the WAL write lock away from the peer.

The TUI stayed up, kept proxying, and silently saved nothing until the process was restarted. Both driver defects are upstream; everything below is the gori-side fix.

## The fix

**1. The writer no longer keeps a connection it cannot let go of.** It checks one out and retires it when a write on it fails (`@db.discard` first, then `close` — a poisoned statement can raise at `sqlite3_finalize`, and a half-closed connection left in the pool is the segfault `#close` already documents). A collision now costs one rolled-back batch, which every caller already handles.

**2. Writer transactions open with `BEGIN IMMEDIATE`.** A deferred BEGIN takes only a read lock and upgrades on its first write — and that upgrade does not go through the busy handler, so SQLite answers `SQLITE_BUSY_SNAPSHOT` immediately and `busy_timeout=5000` never applies. IMMEDIATE takes the lock up front, where the timeout does apply, so two writers queue instead of racing. `Schema.migrate!` already did exactly this and documents the property.

A batch of nothing but index requests skips the transaction entirely — an empty deferred BEGIN/COMMIT was free, an empty IMMEDIATE one is not, and `index_pending!` produces that shape on every `flush`.

**3. `gori mcp --read-only` no longer opens a writable store.** It kept a full writer fiber whose idle FTS indexer ran a write transaction every few milliseconds against the very database the TUI beside it was capturing into. A read-only store starts no writer at all; `@writes` is closed up front so every write API takes the closed-store degradation it already has. The two count-only opens (MCP's `delete_project` dry run and `gori run project delete`'s preview) are read-only now too.

Also in scope, found while testing: `Store#close` let a failing `@db.close` escape, which skipped the open-lock release and left the project answering "open in another gori instance" with the process that held it already gone.

## Verification

`spec/store/concurrent_writer_recovery_spec.cr` — 5 examples, a second connection pool standing in for the peer process (SQLite locks per connection) and `busy_timeout=1` to make the race deterministic instead of a 5-second wait.

Control-run against unpatched `src/`, which is the point:

```
1) capture survives an indexer collision
   Failure/Error: store.insert_flow(capture(3)).should be > 0
     Expected 0 to be GreaterThan 0
```

That is the report, reproduced in 27 ms: the idle indexer collides once, and because capture shares the writer's connection with it, `insert_flow` returns 0 from then on. Two more examples fail unpatched (the FTS backlog never drains; the store cannot even close). All 5 pass with the fix.

Process-level, with the built binary against a database holding 3 dirty rows:

| | `fts_dirty` after the server ran |
|---|---|
| `gori mcp --read-only` | 3 — wrote nothing, held no writer slot |
| `gori mcp` | 0 — indexer still works normally |

And with a peer process holding the WAL write lock across a live `gori mcp`: zero `FTS index batch failed` lines during contention, backlog fully drained after release — the two writers queue now instead of one of them wedging.

Full suite: 9786 examples, 0 failures.

## Notes

- `busy_timeout` stays at 5000 ms. With IMMEDIATE a collision can now block the writer for that long instead of failing instantly, but real transactions here are millisecond-scale, and pre-fix the first write inside the deferred transaction blocked on the same timeout anyway.
- `Schema.migrate!` still runs on a read-only open — a stale schema has to be upgraded to be readable — so "never writes" means after open. The docs say so.

Reported by @d0kk2bi, whose read of the driver behaviour was correct.


hahwul:
Addressed a review pass over this branch. Six of the findings were real; each fix below was control-run against the unpatched tree, so none of them are speculative.

**The one that mattered.** Making the writer take its connection lazily introduced a new raise site: `@db.checkout` raises `DB::PoolTimeout` after five seconds against a full pool, and both index call sites plus `prune` evaluate it in *argument* position — outside the rescues those methods carry for their own failures. The escape killed the writer fiber with the batch's ops already pulled off `@writes`, which puts them out of reach of the drain `writer_loop` does on the way out, so `index_pending!` parked on a reply nobody could send. New spec exhausts a one-connection pool; without the guard it hangs out to its 10 s timeout:

```
1) answers the index request instead of dying with it off the channel
   index_pending! was never answered — the writer died holding its caller's reply
```

**Read-only was still taking one write lock.** `Schema.migrate!` opens `BEGIN IMMEDIATE` before reading `user_version`, so `gori mcp --read-only` blocked at startup against a capturing peer and, on timeout, fell back to "unbound" — failing in exactly the case the mode was added for. It now peeks at `user_version` without the lock and returns when the schema is current; a database this build is ahead of still migrates. Control-run confirms the new spec fails without the fast path.

**A read-only store cannot drain the FTS index**, which made `body:` queries answer from a partial index with no signal. `list_history` / `list_sitemap` now refuse with `FTS_BACKLOG` — the same call the stranded-cursor guard beside them makes, that "no match" and "not indexed yet" must not be the same answer. Verified against the built binary:

```
list_history query:"body:secret"  → isError, "3 flow(s) are not yet indexed … this server is read-only"
list_history query:"host:a.test"  → 3 rows
dirty rows after the server exits → 3   (still writes nothing)
```

Writable servers are unchanged: same query drains 3 → 0 and answers normally.

**A pre-existing index hole, in this PR's path.** `index_pending_batch` SELECTed its rows outside the transaction, so they were a snapshot from before the write lock was held. A peer committing an `update_one` during that wait got its row indexed stale and its `fts_dirty` cleared — permanently unfindable by `body:`, with `fts_backlog` reporting 0. The SELECT moved inside the transaction; an unlocked probe keeps the idle tick from taking the lock when there is nothing to do.

Also: `prune` stops once a sweep has condemned the connection rather than letting the next two burn a `busy_timeout` each; `write_failures` counts `batched` instead of `ops`, so a colliding `flush` no longer reports as lost capture; the spec helpers close the store in `ensure`; and the `--read-only` help text plus `read_only?` docstring, which both overstated the guarantee, are narrowed to what is actually true.

Two findings I did not act on, deliberately:

- **Retiring the connection on *every* write failure**, rather than only when the ROLLBACK was refused. Narrowing it would leave un-reset statements on a connection that goes back to the pool, and finalizing one of those raises out of `Store#close` — costing the last-connection WAL checkpoint. A reopen is well under a millisecond and, with IMMEDIATE, most collisions now wait rather than fail. Documented at the site.
- **`create_project` under `--read-only`** still opens a writable store via `ProjectRegistry#create_or_reopen`, so an agent naming an existing project can touch a database a TUI is capturing into. Real, but it predates this PR and the fix belongs to how project creation is gated, not to the store. This PR strictly shrinks the write surface; I would rather not redesign that path inside a bugfix.

Suite: 9789 examples, 0 failures.

src/gori/store.cr
```diff
@@ -781,18 +832,125 @@ module Gori
...
+    # KNOWN COST, upstream: `DB::Database#discard` is `Pool#delete`, which drops the connection
+    # from the pool WITHOUT the `@availability_channel` nudge `Pool#release` sends. A reader
+    # already parked in `wait_for_available` therefore sleeps out its 5 s `checkout_timeout`
+    # instead of taking the slot this just freed. There is no public API to wake it; the
+    # alternative — releasing a connection we know is bad — is the segfault above.
+    private def retire_writer_conn(conn : DB::Connection) : Nil
+      @writer_conn = nil
+      ::Log.info { "store: discarding the writer's connection after a failed write" }
+      @db.discard(conn)
+      conn.close
```
hahwul: **[bug]** `retire_writer_conn` is the recovery mechanism, but `conn.close` will raise on the very failure this PR recovers from. crystal-sqlite3 resets a statement before `sqlite3_step`, never after a failed one; `sqlite3_finalize` then returns that statement's last error (`SQLITE_BUSY` after a timed-out `BEGIN IMMEDIATE`, a constraint, etc.). `SQLite3::Statement#do_close` `check`s that code and raises. `DB::Connection#do_close` walks the statement cache with `each_value &.close`, so the first poisoned finalize aborts the walk: remaining statements are never finalized, and `SQLite3::Connection#do_close`'s `sqlite3_close` never runs. The rescue at 888–891 logs "leaked, not reused" and continues with a fresh checkout, which is the right trade against the pool-walk SIGSEGV — but a leaked *live sqlite handle* still counts as an open connection. This file already treats that as a defect at the `max_idle_pool_size = 1` comment (a leftover handle skips the last-connection WAL checkpoint; Compact reclaims nothing; copying just the `.db` loses the tail). That used to be a shutdown-only hazard; it is now the happy path of every recovered collision, and a peer that holds the write lock long enough to time out idle FTS will leak another handle per attempt. For `BEGIN IMMEDIATE` BUSY specifically there is also no open transaction, so retrying on the same connection would work: the driver's next `exec` of that statement resets first.

**Suggestion:** Do not retire on a failed `BEGIN IMMEDIATE`. Probe the connection with a read (`SELECT 1`) after the batch rescue; keep it if that works, and only discard+close when a follow-up statement fails or `ROLLBACK` was refused. If a close is still required, finalize/reset statements without `check`ing the deferred error (or `sqlite3_close_v2`) so the C handle actually dies and the checkpoint invariant holds.

src/gori/store.cr
```diff
@@ -1099,15 +1309,18 @@ module Gori
...
         when op = @writes.receive
           return op
         when timeout(tick)
-          # No capture work waiting: index a slice of the backlog and re-check.
-          @fts_backlog_hint = false if index_pending_batch(conn) == 0
+          # No capture work waiting: index a slice of the backlog and re-check. Asking for the
+          # connection each tick (rather than holding one) is what stops an index batch that
+          # failed against a peer process from being retried on the same poisoned connection
+          # every 250 ms for the rest of the session — the shape #752 was reported as.
+          @fts_backlog_hint = false if index_batch_safely == 0
```
hahwul: **[suggestion]** Idle FTS now calls `index_batch_safely` (and therefore `BEGIN IMMEDIATE`) synchronously inside `await_op`'s timeout arm. That wait is not a `select` branch: capture ops that arrive while the busy handler is spinning sit in `@writes`, and `insert_flow` blocks the proxy fiber for up to `busy_timeout` (5 s on `Store.open`). Same-process idle indexing used to fail `SQLITE_BUSY_SNAPSHOT` immediately (then wedge); after this PR a peer holding the lock stalls the data path (P6) for the full timeout, once per slow tick. The empty-backlog `SELECT 1` probe avoids this when there is nothing to index; the reported #752 shape is dirty rows plus a second writer.

**Suggestion:** For the idle tick only, try-lock (`PRAGMA busy_timeout=0` around the idle `BEGIN IMMEDIATE`, then restore) and skip the slice on BUSY, leaving rows dirty. Capture batches can keep the 5 s queue. Also do not treat a failed index (`0`) as an empty backlog when clearing `@fts_backlog_hint` — that is what slows the retry, but it also hides "still dirty, lock was busy."

src/gori/store.cr
```diff
@@ -781,18 +832,125 @@ module Gori
       end
     end
 
+    # The writer's connection, guaranteed usable.
+    #
+    # `using_connection` used to pin ONE connection here for the whole life of the store, and
```
hahwul: **[suggestion]** The new `writer_conn` / `retire_writer_conn` / `write_transaction` / `@read_only` comments (and the MCP `count_project_objects` note at `src/gori/mcp/tools/projects.cr:250`) narrate the change: "used to pin ONE connection", the crystal-db `TopLevelTransaction` postmortem, "HEAD: 0", "what the comment used to say ('read-only-ish') and now is." The non-obvious constraints (why IMMEDIATE, why discard-before-close, why skip empty batches) are worth keeping; the issue-history dump is not.

**Suggestion:** Keep one sentence of WHY per site (IMMEDIATE so the busy handler applies; discard before close to avoid the pool-walk SIGSEGV; read-only closes `@writes` so there is no writer fiber). Move the #752 postmortem to the spec header, which already has it.

Issue #821, opened 2026-08-27
feat(rewriter): response-modification presets (unhide fields, strip validation, drop CSP)


hahwul:
## Summary

Burp ships a set of one-click **response modification** options (unhide hidden form fields, remove `disabled` attributes, remove input length limits, strip JavaScript form validation, remove security headers / CSP). They exist because client-side restrictions are the single most common thing a tester wants gone, and hand-writing the regex for each engagement is busywork.

gori's Rewriter can already express every one of these — it just makes the operator author them from scratch each time. Ship them as **named presets** the Rewriter can install as ordinary rules.

## Motivation

- `src/gori/rules.cr` + the Rewriter tab already do match & replace on response bodies and headers, with scope and enable/disable per rule.
- What is missing is not capability, it is a starting point. A preset list turns a 10-minute regex-authoring detour into two keystrokes.
- Presets stay *ordinary rules* once installed — visible, editable, disable-able. Nothing hidden, nothing magic (P4).

### Preset catalog

New `src/gori/rules/presets.cr` (or `src/gori/rewriter/presets.cr` to match the tab's naming) — a static catalog of `{name, description, rules : Array(rule spec)}`:

| Preset | What it does |
|---|---|
| Unhide hidden form fields | `type="hidden"` → `type="text"` (optionally add a visible border/label) |
| Enable disabled fields | strip `disabled` / `readonly` attributes on form controls |
| Remove input length limits | strip `maxlength=…` |
| Strip client-side validation | remove `pattern=`, `required`, and `on(submit\|change\|input)=` handlers |
| Remove CSP | drop `Content-Security-Policy` / `-Report-Only` response headers |
| Remove security headers | drop `X-Frame-Options`, `X-Content-Type-Options`, `Strict-Transport-Security` (each individually toggleable) |
| Disable SRI | strip `integrity=` from `<script>` / `<link>` |

Each preset is one or more `Store::MatchRule` rows in the shape the Rewriter already persists.

### Surface

- **TUI**: Rewriter tab gains an "Add from preset…" entry (a `PickerOverlay` listing the catalog with descriptions). Selecting one inserts the rules **as normal editable rules**, scoped to the current project, disabled or enabled per the operator's choice at insert time. The existing preview path (`preview_rule` / `src/gori/tui/rewriter_rule_overlay.cr`) should work on them unchanged.
[...]

Issue #772, opened 2026-08-21
TUI: a peer changing probe mode or Match&Replace lands silently, while the Issues-notes change announces


hahwul:
Split out of #771, which made a peer's write reach the live proxy objects. That fix is right, but it landed the *notification* in only one place — and not the consequential one.

## What happens now

After #771, a peer (MCP agent, `gori run …`, a second TUI) can change what THIS session sends on the wire, and the operator is not told:

- **Probe mode.** `set_probe_mode active` from an agent is adopted on the `data_version` tick. That is not a view setting — it is the authorization to start firing attack payloads at the target. The only signal is the `probe:active` chip in the top bar changing, with no toast.
- **Match&Replace / extract rules.** A peer's `create_rule` starts rewriting the bytes of every request and response passing through this session. Verified end to end: a global rule created via MCP after a capture was already running showed up in the traffic that went upstream, with nothing said.

Meanwhile the *least* consequential of the three peer edits does announce:

```
notes changed by another session — esc will refuse to overwrite; ^W discards yours
```

So the tool is loud about a text field and silent about what goes on the wire.

## Why it matters

The whole premise of #771 is a human and an agent working the same project at once. In that setup "the agent turned active probing on while you were driving" is exactly the kind of state change the human needs to see — it is the one with a blast radius outside the tool. The downgrade direction (`off`/`passive`) is safe and needs nothing; the upgrade direction does.

## Shape of a fix

`Runner#notifications` already exists (`notifications.push(:warn, …)`, used for the CA trust-store warning at startup), so there is somewhere to put it. Roughly:

- Announce a probe-mode change adopted from a peer, at least on the UPGRADE to a mode that `probes_actively?`. `Analyzer#apply_stored_mode` already knows both the old and new mode, so it can report which direction it moved.
- Announce a peer's rule set changing the live `Rules`/`Bindings` snapshot — coalesced, not per-rule, since the tick fires on this session's own captures too. `Colormarker#refresh` already bails out on an unchanged rule set, so there is a precedent for a cheap "did it actually move" answer to gate this on.
- Keep it latched the way the Issues notes announcement is, so a live capture does not print a line a second.

Deciding the exact policy (which changes are worth a line, warn vs info, whether the headless `gori run capture` should log the same thing) is the substance of this issue — hence splitting it out of #771 rather than bolting it on.

PR #838, opened 2026-08-28
feat(rewriter,decoder,probe): external process hooks — pipe bytes through an operator's own command (#818)

### LINKED ISSUE #818: feat(rewriter,decoder,probe): external process hooks — pipe bytes through an operator's own command
## Summary

gori has no way to plug an **arbitrary transform** into the data path without recompiling Crystal. Burp solves this with BApps/Bambdas, Caido with JS plugins and node-based Workflows. gori's Rewriter is static match & replace, custom Probe rules are regex-only, and the Decoder catalog is a fixed set of built-in converters. The moment an operator needs "run these bytes through *my* tool", they are stuck.

The gori-native answer is **not** a JS plugin SDK — it is the UNIX pipe. Bytes in on stdin, replacement bytes out on stdout. Language-agnostic, reuses every script the operator already owns, and keeps P7 (raw bytes are the truth) intact.

## Motivation

- Rewriter (`src/gori/rules.cr`, a `Proxy::HeadRewriter`) can match/replace literal or regex, but cannot *compute* a replacement (re-sign a JWT, recompress a body, recalculate an HMAC, decrypt a proprietary envelope).
- Decoder chains (`src/gori/decoder.cr`) are limited to the built-in converter catalog.
- Custom Probe rules (`src/gori/probe/custom_rule.cr`) can only regex-match; they cannot shell out to a real detector.

A single "shell out to argv" primitive covers all three surfaces.

## Design

Introduce one shared primitive and wire it into three existing seams.

### Core primitive: `Gori::ProcessHook`

New file `src/gori/process_hook.cr`. A pure, surface-neutral runner:

```
Gori::ProcessHook.run(argv : Array(String), stdin : Bytes,
                      timeout : Time::Span, env : Hash(String,String)) : Result
# Result = {status: Int32, stdout: Bytes, stderr: String, timed_out: Bool}
```

Contract (these are the load-bearing invariants — call them out in comments):

1. **P6 — never stall the data path.** A hard timeout (default 5s, configurable) is mandatory. On timeout / non-zero exit / spawn failure, the **original bytes pass through unchanged** and the failure is surfaced as a non-fatal notice — never a dropped or hung flow. The proxy path must remain responsive even if the hook wedges.
2. **Bounded output.** Cap stdout (e.g. 32 MB, same order as body caps elsewhere) so a runaway child cannot exhaust memory.
3. **No shell.** `argv` is exec'd directly (`Process.new(argv[0], argv[1..]`, `shell: false`) — no `sh -c`, so operator payloads never get shell-interpreted.
4. **Explicit opt-in + visible.** Because a hook runs arbitrary local commands, it is configuration the operator wrote (P4), and each configured hook is listed plainly in settings so it is never invisible.

### Seam 1 — Rewriter action `pipe`

`src/gori/rules.cr` / `src/gori/store` match-rule model gains a new op kind `pipe` whose "replacement" field is the argv. When a rule matches, the matched region (or whole body — decide per `RuleTarget`) is fed to `ProcessHook.run`; stdout replaces it. Env carries `GORI_FLOW_METHOD`, `GORI_FLOW_URL`, `GORI_FLOW_HOST`, content-type, etc. so the hook has context.

- TUI editor: `src/gori/tui/rewriter_rule_overlay.cr` gets the new op in its option cycle.
- Note the two-line-view hazard (`memory: two-line-views-one-message`) — a body-region pipe on the request path must respect the same CRLF view the parser saw.

### Seam 2 — Decoder chain step `| argv`

`src/gori/decoder.cr` chain grammar (`base64-decode > url-encode`) gains a pipe step: `base64-decode > | ./mytool --flag > json-pretty`. Autocomplete (`src/gori/tui/chain_pane.cr` / `decoder_view.cr` `ChainComplete`) leaves `|` steps as free text. This makes the Decoder tab and Fuzzer `¦chain` markers scriptable.

### Seam 3 — Probe rule kind `exec`

`src/gori/probe/custom_rule.cr` gains an `exec` kind: response bytes → argv → exit code / stdout decides whether a finding is raised (e.g. exit 0 = match, stdout = evidence line). Passive by default; active-mode gating unchanged.

## Where to implement

| Piece | File |
|---|---|
| Core runner | `src/gori/process_hook.cr` (new) |
| Rewriter op | `src/gori/rules.cr`, `src/gori/store/*` (rule model), `src/gori/tui/rewriter_rule_overlay.cr` |
| Decoder step | `src/gori/decoder.cr`, `src/gori/tui/chain_pane.cr`, `src/gori/tui/decoder_view.cr` |
| Probe kind | `src/gori/probe/custom_rule.cr`, `src/gori/tui/custom_rule_overlay.cr` |
| CLI/MCP parity | `src/gori/cli/run/rewriter.cr`, `src/gori/verbs/rewriter.cr`, corresponding MCP tools |

## Design principles to honor

- **P6**: timeout + pass-through-on-failure is non-negotiable on the proxy path.
- **P7**: the hook sees and returns raw bytes; gori does not re-encode around it.
- **P4**: hooks are operator-authored config, listed visibly; gori never invents a hook.
- **P1/§2**: the runner is one engine; TUI/CLI/MCP all call it, none re-implement.

## Acceptance criteria

- [ ] A Rewriter `pipe` rule transforms a matching request/response body via an external command, live in the proxy.
- [ ] A hook that exits non-zero, times out, or fails to spawn leaves the flow **unchanged** and posts a notice; the proxy stays responsive.
- [ ] Decoder chain accepts a `| argv` step in TUI and CLI.
- [ ] A Probe `exec` custom rule raises a finding based on an external command's result.
- [ ] `argv` is exec'd without a shell; specs cover the timeout, non-zero-exit, and oversized-output paths.

## Scope notes

- This is **not** a plugin API with a stable ABI, a JS runtime, or in-process extensions. Just process boundaries. Keep it that way (P0).
- Out of scope: sandboxing the child process. Document that hooks run with the operator's privileges — same trust level as `--config` or a Rewriter rule already has.

---

### 한글 요약

gori는 재컴파일 없이 **임의 변환**을 데이터 경로에 끼워 넣을 방법이 없다(Burp의 BApps/Bambdas, Caido의 플러그인/Workflows에 대응하는 축이 통째로 비어 있음). gori다운 답은 JS 플러그인 SDK가 아니라 **UNIX 파이프** — bytes를 stdin으로 주고 stdout을 교체 bytes로 받는 `Gori::ProcessHook` 하나를 만들어 세 군데에 연결한다: (1) Rewriter의 `pipe` 액션, (2) Decoder 체인의 `| argv` 스텝, (3) Probe의 `exec` 룰. 핵심 계약은 **P6** — 하드 타임아웃 + 실패 시 원본 통과(프록시 경로가 절대 멈추지 않음)와, 셸 없이 argv 직접 실행. 언어 무관·기존 스크립트 재사용·P7(raw bytes) 유지가 장점.

hahwul:
Closes #818.

Pipe bytes through the operator's own command. One primitive — `Gori::ProcessHook` (`src/gori/process_hook.cr`) — wired into three existing seams, none of which re-implements it (P1/§2):

| Seam | Where | What it does |
|---|---|---|
| Rewriter `pipe` op | `rules.cr`, TUI editor, `gori run rewriter --op=pipe`, MCP `create_rule` | The matched region goes to the command; its stdout replaces it, live in the proxy |
| Decoder `exec:` step | `decoder/chain.cr`, Decoder tab, `gori run decoder`, `§value¦chain§` markers | One chain step is a command instead of a converter |
| Probe `exec` rule | `probe/custom_rule.cr`, Probe rules, `gori run probe rules add --exec` | The region goes to the command; exit 0 raises a finding, stdout is the evidence |

This is gori's answer to BApps/Bambdas and JS plugins, and the deliberate choice is a UNIX process boundary rather than an embedded runtime: language-agnostic, reuses every script the operator already owns, no stable ABI, no recompile (P0).

## The P6 contract, and the failure paths

This is the part worth reviewing.

`ProcessHook.run` **returns within `timeout + KILL_GRACE + 2 × COLLECT_GRACE` under every failure mode it can reach**, and the timeout is enforced there rather than by each caller's good intentions. It SIGKILLs a child that overruns, and abandons a pipe that will not close rather than waiting on it — every channel is buffered to capacity 1 precisely so an abandoned pump's `send` completes and its fiber ends instead of parking on a receiver that already returned.

It **never raises**. A spawn failure (ENOENT, EACCES, EISDIR, a fork that fails) comes back as `Result#spawn_error`, because a raise on the proxy path would drop the operator's flow over a typo in a rule.

Five things can go wrong, and all five have the same disposition at the rewrite seam — **the original bytes pass through unchanged, and the reason is written to the event feed as a non-fatal notice**:

| Failure | Result |
|---|---|
| argv does not tokenize | refused at the write surface; at apply time, bytes unchanged + notice |
| command cannot be spawned | bytes unchanged + notice |
| hard timeout | child SIGKILLed, bytes unchanged + notice |
| non-zero exit | bytes unchanged + notice (with the child's stderr tail) |
| stdout past 32 MiB | child killed, bytes unchanged + notice |

There is no "fail closed" option worth having here: refusing the message would take the operator's traffic down over their own script, and splicing in a partial or empty replacement would put corrupt bytes on the wire (P7).

**The Rewriter's timeout is a BUDGET, not a per-spawn limit.** A per-spawn timeout bounds one child; a regex matching four hundred times in a body would then multiply it by four hundred while the client waits. One rule gets one budget for one message, handed to each match as the remaining time; once it is gone the remaining matches pass through unchanged with one notice. `spec/rules/pipe_spec.cr` proves it with twenty matches of `sleep 30` against a 1s budget.

Verified live, not just in specs — a `pipe` rule uppercasing `world` in a response body through `gori run capture`, alongside a second rule pointed at `/nonexistent` on the same body:

```
$ curl -s -x http://127.0.0.1:18096 http://127.0.0.1:18095/index.txt
hello WORLD SECRET here
$ sqlite3 gori.db 'select kind, message from events order by id desc limit 1'
hook_failed|rewrite rule "broken" pipe failed: /nope/missing: could not run it
             ('/nope/missing': No such file or directory) (bytes passed through unchanged)
```

## No shell

`argv` is exec'd directly (`Process.new(cmd, args, shell: false)`). `ProcessHook.parse_argv` honours whitespace splitting, `'…'`, `"…"` with `\"`/`\\`, and a bare `\` — **and nothing else**. `$FOO`, `*`, `` ` ``, `;`, `&&`, `|` and `>` are ordinary characters that land in an argv element verbatim, so captured bytes flowing through a hook's arguments can never be shell-interpreted. A NUL in any element is refused: `execvp` would truncate the argument there and run a different command than the one the operator read back in the editor.

## Notes for the reviewer

**`RuleOp#header?` was a negation, and that was a trap.** It read `!replace? && !short_circuit?` — "everything that is not one of the two ops I know about" — so adding `Pipe` silently made it a header op: `normalize_shape` would have forced every pipe rule onto the head, the `ws`/`body` refusals would have fired on legal shapes, and `apply` would have routed its bytes into `head_set_header`. It is now an explicit list.

**The `| argv` syntax in the issue could not be used, and the code says why.** `|` is already one of the Decoder grammar's three interchangeable separators (documented in the CLI reference), so `a > | ./tool` splits into a bare `./tool` token with the marker gone before anything can read it — and repurposing `|` would silently change what every existing `a | b` chain means, from "run converter b" to "exec the program b". The marker is `exec:`, the colon-prefixed kind idiom `DisplayColumns` already uses. The cost is inherent to the flat grammar: `>`, `|` and `,` cannot appear in an `exec:` step's arguments, which is documented. `a | b` is unchanged.

**A hook failure at the Decoder seam FAILS the step instead of passing the input through**, unlike the Rewriter. The two seams answer to different halves of the same principle: the Rewriter is on the proxy data path where P6 says a broken hook must never cost a flow, while the Decoder is an interactive workbench where carrying the input forward as if the step had run would hand the operator a value that is not what the chain says it is. A `Failed` row with the reason is what every other converter failure already gives.

**A Probe `exec` hook that cannot RUN raises nothing and reports.** Treating a spawn failure as a hit would fill Issues from a typo; treating it as a clean no-match and saying nothing is the "absence of a finding reads as clean" shape — the operator cannot tell "the detector looked and found nothing" from "the detector never ran". A non-zero exit is neither: that is the rule's own answer and is silent by design. Failures are de-duplicated per rule for the life of the rule list, so a scan over ten thousand flows with a mistyped command writes one event row.

**The per-frame Rewriter preview does not run hooks.** `sync_preview_out` is called from `render_rules` — a `pipe` rule there would fork the operator's command sixty times a second over a sample they are merely looking at, and a hook is by definition allowed to have side effects. `transform_message` grew a `run_hooks` flag (default true, because the other caller — `mcp/tools/send.cr` — is a real send); the pane says which rules it left out. Separately, `rule_affects?` counts a pipe rule by whether its PATTERN matches rather than by whether the transform changes the bytes: `preview` replays a candidate over 500 recent flows on the rule editor's keystroke path, and "does the output differ" would mean spawning the command once per flow.

**P7's limit at the Probe seam is called out in the code.** An `exec` rule sees the region text `Passive::Context` hands every custom rule — content-decoded, capped, scrubbed — not raw octets, because a rule table where `exec` and `regex` looked at different bytes for the same {side, region} would make two rules over one flow mean two different things. A hook needing the exact wire bytes has `GORI_FLOW_ID`.

## The #772 question: does a peer's `pipe` rule need a stronger signal?

**Yes, and it is implemented.** The gap is a category one, not a volume one.

The existing announce already says the right thing about a Match&Replace rule — a peer changed what THIS session puts on the wire — but every word of it is about **bytes**: worst case a peer rewrote a header on traffic you are watching, and the bytes are on screen in History either way. A `pipe` rule is not that. Adopting it means this process will fork and exec a command off local disk, with this operator's privileges, every time a message matches, and none of that is visible in any pane, because what shows up in History is only the OUTPUT. *"1 Match&Replace rule changed by an agent — rewriting live traffic here"* is a true sentence that would not have told them.

So `RuleSetChange` grows an `executes` field (its own field for the same reason `reordered` is one — it is not a count of anything the other numbers describe), and `PeerNotices#rule_notice` pre-empts every other wording with:

> **1 Match&Replace pipe rule added or changed by an agent — it runs a local command here, with your privileges, on every message it matches**

Always `:warn`. Three deliberate narrowings:

- It counts **changed** entries, not the standing total. A peer editing an unrelated rule while a pipe rule has been sitting there for an hour is not news about the pipe rule, and repeating the loud line every time anything moves is how a loud line stops being read.
- It counts **enabled** ones only. A disabled pipe rule forks nothing; it is a row, not a hook.
- It rides the same trailing-edge quiet window and survives coalescing, so a burst that includes one pipe rule still announces once — as the loud line.

## Scope

`hooks.timeout_secs` is a new settings scalar (one knob for all three seams, clamped 1–60; the ceiling is not negotiable by configuration). Sandboxing the child is explicitly out of scope and documented as such: hooks run with the operator's privileges, the same trust level a Rewriter rule or `--config` already has. #821 (response-modification presets) is untouched.

## Verification

- `crystal build --no-codegen src/main.cr` ✅
- `crystal spec` — 11255 examples, 0 failures ✅
- `./scripts/bench_check.sh` — 48 harnesses ✅
- `crystal build --no-codegen scripts/seed_demo.cr` ✅
- `crystal tool format` clean; `ameba` on the changed files adds no new findings beyond the repo's pre-existing complexity/`not_nil!` noise (one 13/12 on the argv tokenizer, which is a character state machine)
- 42 new examples across `spec/process_hook_spec.cr`, `spec/rules/pipe_spec.cr`, `spec/decoder_exec_spec.cr`, `spec/probe/exec_rule_spec.cr`, plus three in `spec/peer_notices_spec.cr` — the timeout, non-zero-exit, spawn-failure and oversized-output paths each have their own, and the no-shell property has one that asserts metacharacters survive as ordinary argv elements
- End-to-end through the real binary: all three seams, plus a live proxy rewrite and its failure notice (transcript above)

## Acceptance criteria

- [x] A Rewriter `pipe` rule transforms a matching request/response body via an external command, live in the proxy
- [x] A hook that exits non-zero, times out, or fails to spawn leaves the flow unchanged and posts a notice; the proxy stays responsive
- [x] Decoder chain accepts an external-command step in TUI and CLI (`exec:` rather than `|` — see above)
- [x] A Probe `exec` custom rule raises a finding based on an external command's result
- [x] `argv` is exec'd without a shell; specs cover the timeout, non-zero-exit, and oversized-output paths


hahwul:
## Review round 1 — findings addressed (`f50485e4`)

A `/code-review` pass over the branch found four defects that mattered and a set of smaller ones. All are fixed with regression coverage; `crystal spec` is 11263 examples / 0 failures, bench + seed_demo build.

### The four that mattered

**1. MCP `decode` was an ungated path to local code execution.** `decode` sits in `UNBOUND_SAFE`, is not in `AGENT_ACTION_TOOLS`, and is never checked against `@allow_actions` — it is reachable from `gori mcp --read-only` with no project bound, and documented "pure: no store, no network". Adding `exec:` to the chain grammar turned it into RCE: `exec:/bin/sh -c …` needs no shell, because `/bin/sh` *is* the argv. It now refuses. The refusal asks the **registry**, not the spec text, because a saved chain is callable by name — `myenc` can carry an `exec:` step with nothing in the token to say so; `Converter#runs_commands?` carries that fact from `Library.build`. The other two seams stay as they were: `create_rule op=pipe` and `create_probe_rule match_kind=exec` are gated writes that log an agent action.

**2. A lost stdout pump read as a successful empty run.** `collect` returned a fallback value, so a hook whose stdout pipe was still held open past `COLLECT_GRACE` (a helper it backgrounded inherits the fd; the direct child exits 0) came back with `ok? == true` and empty `stdout` — and `pipe_apply` spliced `""` over the matched region on the live wire, silently. `collect` now answers `nil`, and `Result#output_lost` makes it a failure with its own sentence. This was the worst bug on the branch: it broke the one invariant the whole design rests on.

**3. `$KEY` in a pipe rule's argv was an argument-injection primitive.** `replacement_for` expanded bindings across the whole string and `parse_argv` tokenized the result, so an origin that mints `TOKEN` as `x --config /tmp/evil.yml` turned `./resign.sh --key $TOKEN` into five arguments. `head_scoped?` returns false for pipe (correctly — CR/LF forge nothing in an argv), so nothing caught it. Fixed by inverting the order: `pipe_argv` tokenizes the literal spec **first** and resolves each `$KEY` inside the element it sits in, so a value can never add, split or remove an argument however it is spelled. There is no shell here, so quoting could not have saved it; the split has to not happen at all. Regression test asserts three arguments, not five.

**4. A backslash inside `"…"` escaped any character and was deleted.** `"a\d+"` tokenized to `ad+`; `"C:\tools\dev.pem"` to `C:toolsdev.pem`. The rule validated at every write surface and then ran a command the operator never read back. It now escapes only `"` and `\`, as a POSIX shell does — and my own doc comment two lines up had said so all along.

### The rest

| Finding | Fix |
|---|---|
| Budget was per-RULE (N rules × 60s on one head) | One deadline per `apply` call, shared by every pipe rule in it; code, docs and this PR now state the bound a *message* sees (two rewrites, head and body) |
| Probe `exec` with an unparseable argv: silent no-match | Reports. `Settings.parse_scan_rules` clamps a hand-edited settings.json without calling `valid_pattern?`, and a hand-edited global rule is a supported way to write one — so this was the "absence of finding reads as clean" failure the method's own comment argues against |
| Notices de-duplicated on a sentence carrying the child's stderr | `Result#reason` / `failure_key` split out; a hook printing a timestamp before failing was writing one SQLite row per proxied message *and* growing an unbounded `Set` for the length of a scan |
| PeerNotices' pipe branch discarded a concurrent extract change | `flush` has already cleared both, so it could never be re-announced; the extract fact now rides on the end of the pipe line |
| Bare `exec:` fell through to the registry | `exec_spec` accepts an empty remainder; the step fails with `no command` instead of `unknown converter "exec:"` |
| `pipes_for?` ignored `part` | Goes through `rewrites?` like every other liveness gate; a `part: ws` rule no longer makes the HTTP preview claim it skipped something |
| argv tokenized + env Hash built per message before knowing the pattern matched | env built lazily on the first actual run |
| Probe `exec` runs once per flow on the analyzer fiber | Documented where the cost is paid, plus in the guide — a slow detector is the whole rule set's bottleneck under live capture, and `Store#publish` drops onto `@probe_events` when it backs up |
| Orphaned `normalize_shape` doc; a stale claim that the preview runs the hook | Both fixed |
| No CHANGELOG entry | Added |

### One finding kept as-is, deliberately

`Result#failure` puts up to 200 chars of the child's stderr into the event row, which `command_label`'s argv[0]-only redaction does not cover — a hook that echoes its stdin on failure echoes it there. Kept: it is the only part that makes a notice actionable ("exited 4" alone sends the operator to their own logs), it is the operator's own script on their own machine, and the row lands in a database that already holds the full captured traffic. What changed is the comment, which claimed a property the pipeline does not have end to end. `command_label` now says it bounds what *gori* writes, not what ends up in the row.

Issue #824, opened 2026-08-27
feat(diff): retest workflow — diff two projects/snapshots at endpoint scale


hahwul:
## Summary

Retesting is roughly half of real engagement work — "what changed since last time?" — and every tool on the market is weak at it. Burp, Caido, and gori all diff *two messages*; none diffs *two engagements*. gori's Comparer already handles the flow-vs-flow axis (`src/gori/tui/comparer_view.cr`). This issue lifts that axis up to **project / sitemap scale**:

- Which endpoints are **new** since the last assessment?
- Which endpoints **disappeared**?
- Which endpoints now answer **differently** (status class, size band, content-type, auth behavior)?
- Which previously-filed **issues** still reproduce?

gori is structurally well-suited: projects, sitemap, issues, and flows all live in one store per project (`src/gori/project.cr`, `src/gori/project_registry.cr`, `src/gori/store.cr`), so "compare project A to project B" is a real, cheap query rather than an import/export dance.

## Motivation

- Today the only way to answer "is this endpoint new?" is memory or a manual export diff.
- The retest deliverable (a delta report) currently has to be assembled by hand.
- gori's `Sitemap` is already a normalized, folded tree (`fold_templates!`, `fold_queries!`, `group_sequences!`) — meaning `/user/1` and `/user/2` already collapse to one node. That normalization is exactly what makes a *meaningful* diff possible instead of a noisy one.

### 1. Comparison unit

Compare two **snapshots**. A snapshot is either:

- another **project** (the common case: last quarter's engagement vs this one), or
- a **point in time within one project** (flows before/after a timestamp), or
- a saved, explicit **baseline** captured on demand.

Start with **project-to-project** — it needs no new persistence and answers the main question. A stored baseline can follow.

### 2. Endpoint identity (the crux)

Diff quality lives or dies on what counts as "the same endpoint". Reuse `Sitemap`'s existing folding so ids/dates/sequences are already collapsed:

- key = `{host, method, folded path template}` (`Sitemap.normalize_path`, `template_class`, the fold passes)
- record per key: status classes seen, size band, content-type, auth-required signal
- **Do not** invent a second normalization — reuse `src/gori/sitemap.cr`. A key that disagrees with the sitemap tree would make the diff disagree with the tab the operator is looking at.

Beware `Node#path` holding a full path string (see `memory: sitemap-depth-quadratic`) — build keys without repeatedly materializing full paths.

### 3. Diff result

```
added:     endpoints present in B, absent in A
[...]

PR #850, opened 2026-08-28
feat(diff): a retest row reaches Issues and Notes (#845)

### LINKED ISSUE #845: feat(diff): a retest row should reach Issues and Notes
## Summary

The retest diff (#824) finds the right things and then drops you. Its complete action set:

```
src/gori/tui/runner/diff.cr
  diff_pick  diff_swap  diff_run  diff_cycle_lens  diff_move  diff_rows_shown?  diff_to_comparer
```

You can hand a row to the Comparer. There is **no path to Issues or Notes** — `grep -n "issue" src/gori/tui/controllers/diff_controller.cr` returns one hit, and it is a comment about quadratic memory.

The deliverable of a retest is a list of findings. The diff produces exactly the input to that list — "this endpoint answered 403 last quarter and answers 200 now" — and then makes the operator retype it somewhere else.

## Why this belongs to Diff and not to the operator

Every other finding-producing surface in gori already has this exit. Probe promotes to an Issue. Discover's results open a flow. The Comparer reaches the flows it is diffing. Diff is the one surface that produces a judgement and cannot record it, and it is the surface whose *entire output* is a list of things worth writing down.

There is also information here that only Diff has, and that a retyped issue loses: which two projects were compared, which lens produced the row, and what the two sides actually were.

## Design

Add the exit, carrying the context the row already holds.

- **Create an Issue from a row.** Prefill from the row: the endpoint (host + method + folded path template), what changed (status class / size band / content-type / auth signal, whichever the lens flagged), and both sides' identity. Reuse `IssueForm` — do not build a second issue editor.
- **Link, do not copy.** `entity_links` (`src/gori/links.cr`) is how gori already relates an issue to the evidence behind it. An issue created from a diff row should link to the flow on each side where those exist, so the issue can be opened later and still reach the bytes.
- **A note, too.** The lighter exit matters for retest: most rows are "worth mentioning", not "a finding". `create_note` from a row should be one keystroke.
- **Say what is being recorded when it is a coverage gap.** #824 was explicit that an endpoint absent from the newer side may mean "removed" or "never visited this time", and that the two must stay distinguishable. An issue created from such a row must carry that distinction into its text rather than asserting a removal the diff did not observe.

Reach all three surfaces where it makes sense: `gori run diff --format md` already produces a report section, so consider whether the CLI/MCP path wants "create issues from this diff" as a batch, or whether that is a TUI-only affordance. Argue the choice in the PR — a batch that files fifty issues from a noisy diff is worse than no batch at all.

## Where

| Piece | File |
|---|---|
| Verbs + keymap | `src/gori/tui/runner/diff.cr`, `src/gori/verb/context/` (a diff context), `src/gori/verb/keymap.cr` |
| Row → issue/note payload | `src/gori/tui/controllers/diff_controller.cr` |
| Issue editor (reuse) | `src/gori/tui/issue_form.cr` |
| Links (reuse) | `src/gori/links.cr`, `entity_links` |
| Diff model | `src/gori/diff/` |

## Design principles to honor

- **P1**: the Issue is created through the same path `create_issue` already uses; the diff does not learn to write issue rows.
- **P3**: reuse `IssueForm` and `entity_links` rather than inventing a diff-specific record.
- **Honest absence**: a coverage-gap row must not become an issue that claims an endpoint was removed.

## Acceptance criteria

- [ ] A diff row can create an Issue, prefilled with the endpoint and what changed, without leaving the tab.
- [ ] A diff row can create a Note in one keystroke.
- [ ] The created Issue links to the flows behind both sides where they exist, and opening it later reaches them.
- [ ] A row that represents a coverage gap ("not seen on the newer side") produces text that says so, not "removed".
- [ ] The two compared projects are identifiable from the created Issue.
- [ ] Existing diff behaviour and specs are unchanged.

## Scope

Not in scope: changing how the diff decides what changed, or its tolerance model. This issue is only about the exit.

---

### 한글 요약

리테스트 diff(#824)는 옳은 것을 찾아놓고 손을 놓습니다. `runner/diff.cr`의 액션 전부가 `diff_pick`/`diff_swap`/`diff_run`/`diff_cycle_lens`/`diff_move`/`diff_to_comparer`이고, Comparer로는 넘길 수 있지만 **Issues·Notes로 가는 길이 없습니다**. 리테스트의 산출물은 발견 목록인데, diff는 그 입력("지난 분기엔 403이던 게 지금 200")을 만들어놓고 운영자에게 딴 데다 다시 치라고 합니다.

**이건 Diff의 몫입니다.** gori의 다른 발견 생성 표면은 전부 이 출구가 있습니다 — Probe는 Issue로 승격하고, Discover 결과는 flow를 열고, Comparer는 비교 중인 flow에 닿습니다. Diff만 판단을 만들어놓고 기록을 못 하는데, 하필 **출력 전체가 적어둘 가치가 있는 것들의 목록**인 표면입니다. 게다가 Diff만 아는 정보(어느 두 프로젝트를 비교했는지, 어느 렌즈가 이 행을 띄웠는지, 양쪽이 실제로 뭐였는지)가 손으로 옮기면 사라집니다.

설계: 행이 이미 들고 있는 맥락을 실어 출구를 추가. Issue 생성은 `IssueForm` 재사용(두 번째 이슈 편집기 금지), 증거는 복사가 아니라 `entity_links`로 **연결**(나중에 열었을 때 바이트에 닿게), Note는 한 키(대부분의 행은 "발견"이 아니라 "언급할 만함"이라 리테스트에선 이 가벼운 출구가 더 중요). **커버리지 갭 행은 조심** — #824가 "새 쪽에 없음"이 *삭제됨*인지 *이번엔 안 가봄*인지 구분돼야 한다고 못 박았으니, 거기서 만든 이슈가 관측하지도 않은 삭제를 주장하면 안 됩니다. CLI/MCP에 일괄 생성을 둘지는 PR에서 논증할 것 — 시끄러운 diff에서 이슈 50개를 쏟아내는 배치는 배치가 없느니만 못합니다.

hahwul:
Closes #845.

The retest diff found the right things and then dropped you. Its whole action set was `diff_pick` / `diff_swap` / `diff_run` / `diff_cycle_lens` / `diff_move` / `diff_to_comparer` — you could hand a row to the Comparer, and that was the exit. The deliverable of a retest is a list of findings; this tab produces exactly that list's input and then made the operator retype it, losing the part only Diff holds.

`⇧F` files the selected endpoint as an **Issue**. `n` records the same text as a **Note** in one keystroke, no form. Neither moves you off the tab.

## The sentence this had to get right

#824 was explicit that an endpoint absent from the newer side means *removed* **or** *never visited this time*, and that the two must stay distinguishable. A record is where a verdict stops being a screenful an operator re-reads and becomes a claim in a client deliverable, so that is where the split either survives or quietly dies.

A coverage-gap row files as `info` — not a finding, which is the whole reason `removed` and `gone` are two verdicts — and says:

```
Verdict    Not seen in B — no request captured (coverage gap, not proof of removal)
…
q3 captured NO request to this endpoint, so this diff observed nothing about it. That
is a gap in the retest's coverage, not evidence the endpoint was removed — request it
before recording it as gone.
```

Every occurrence of the word in that record is a denial of it, and `spec/diff_spec.cr` pins the count exactly (3) so a future edit that adds a bare "removed" anywhere in the text fails rather than ships.

`added` gets the mirror caveat one side over, which the report itself did not have: absence from the **baseline** is equally ambiguous, so the record says "it may be new, or the baseline may simply never have visited it".

## Link, do not copy — and where the link cannot reach

`entity_links.ref_id` is a bare rowid with no project column, and the TUI holds a writer on exactly one store: the open project. So the linkable side is whichever slot names it (B by default, A after `s`), and it is linked as evidence — `insert_issue`'s own `entity_links` write, so opening the issue later still reaches the bytes. The other side's flow is **named** with the target it was captured on and its database path, because a diff-shaped issue that silently dropped half its evidence would be the copy this feature exists to replace.

Two more facts the record keeps apart that a single flag would have merged:

- **pruned vs. cross-project.** A capture deleted between the run (`r` takes a snapshot) and the keystroke is dropped rather than filed against a rowid SQLite will hand to an unrelated flow later — and it is worded as "no longer captured in this project", not as "in that project's own database", which would send the reader somewhere it was never going to be.
- **which `q3`.** The project picker already prints a date beside each name because two engagements against one target are routinely named alike. A record whose only handle is the word "q3" identifies neither side, so both database paths ride along under `Projects`.

## Shape

`Diff::Record` (`src/gori/diff/record.cr`) is pure — no store, no TUI, no clock — and is the single place a row becomes text. The Issue body, the Note body and the JSON row all read it, so they cannot describe one row three ways. `Render.facts_line` went public rather than being respelled for the same reason.

Creation goes through the paths that already exist (P1/P3): `IssueForm` + `create_issue_from_form` for the issue, `Notes`/`NotesController` for the note, `commit_link_to_owner` for the link. The diff learned to *draft*, not to write issue rows.

One shared-code change, in `IssueForm`: a `stay_on_create` flag. `create_issue_from_form` already had two landings — jump to Issues (History's `⇧F`: file it and go look at it) and stay (a workbench picker mid-recon) — and `link_ref` was doubling as the predicate for which. That is not the same question: a diff row whose capture lives in the other project's database has nothing to link and still must not move the operator off the list. The flag names the real predicate; the existing branches are untouched.

`NotesView#new_note` gained an optional `text` (`duplicate_current` already built a note from a string that way) and `NotesController#create_note_with_text` returns `{id, persisted?}` so the toast can say "NOT saved yet (project busy)" rather than claim a write the store refused.

## The judgement call: no CLI/MCP batch

**No batch, and no new CLI/MCP surface.** Four reasons:

1. **The volume is the argument.** `Render::LISTED` already drops `unchanged` from the default listing because on a real retest it is most of the rows. A batch over what is left is "file fifty issues from a noisy diff", which the issue itself names as worse than no batch.
2. **An issue is a judgement; the diff produces candidates.** Auto-filing is inferring a finding the diff did not make (P4). The TUI gesture is per-row *because* the operator is reading the row.
3. **`--format md` is already the batch answer**, and the right one: one section, pasted into the deliverable, reviewed as a whole rather than as fifty rows someone now has to close.
4. **A batch has nowhere honest to put the links.** `gori run diff --from A --to B` may name neither the current project nor any project at all, and the evidence link is the half of this feature that is not retyping.

What the headless path *was* missing is the sentence, not the loop. An agent reading `endpoints` and calling `create_issue` per row never reads the report-level `caveats`, so `--format json` now carries `observation` on every row — verbatim the sentence the TUI's Issue carries. `gori run diff` and MCP `diff_projects` both go through `Render.json`, so both get it. Filing stays one deliberate `create_issue` + `add_link` per row, which is the granularity that was already right.

## Verification

`crystal build --no-codegen src/main.cr` · `crystal spec` (**11,476 examples, 0 failures**) · `scripts/bench_check.sh` (48 harnesses) · `crystal build --no-codegen scripts/seed_demo.cr` · `crystal tool format` · ameba clean on every new file, no regression on the touched ones.

Two spec files, 21 examples: the pure record (the coverage-gap wording and its exact word count, the mirror caveat on `added`, severity, the multi-axis title, the two unlinkable cases, the note's first line) and the controller against a real `Session` plus a second project on disk (which side is linked, that it follows the open project to slot A after `s`, that a pruned capture is refused, and the coverage-gap wording end to end off a real read) — plus the verb registration, its gate, and the space-menu key collision that pushed `diff.issue` off the `'a'` its siblings on other tabs use.

Driven end to end through a tmux pty on two seeded projects: `⇧F` on a `changed` row files with the capture attached and leaves the cursor where it was; `n` on the coverage-gap row reports `note filed` (nothing to link) and on the changed row `note filed · capture linked`; `gori run issues --format markdown` prints the record with its linked flow's request and response underneath.


hahwul:
Review pass (`/code-review` at xhigh, 14 findings). Nine acted on, four rejected with reasoning, one restated.

## Fixed

**Scrub at the chokepoint, not on the parts.** `Record.title`/`body` scrubbed the endpoint and the project labels — but the endpoint is not the only captured string on those lines. `MediaType.essence` lower-cases a `Content-Type` and drops its parameters without touching control bytes, so a moved `content-type` axis carried the response's own header into the IssueForm title through `Change#to_s`, and into the body through `Render.facts_line`. Confirmed against a real capture: `select quote(cast(content_type as blob))` → `X'746578742F68746D6C1B5D303B70776E07'` — ESC (`1B`) and BEL (`07`) stored verbatim, exactly as P7 requires. Now one `one_line` on the title, one `scrub_controls` on the body (a document — its newlines are structure), one on the host. Driven through the TUI end to end with that capture; nothing with a control byte reaches `issues.title`, `issues.notes` or `notes.docs`.

**A note's body claimed a link before the link existed.** An issue is one transaction — `insert_issue` writes the row and its `entity_links` row together, and the shell bails on `new_id == 0` — so the issue body may say "linked to this record" up front. A note is *two* writes, and building the text first let a store-busy rollback leave a note asserting evidence that was never attached: a false evidence claim in something an operator pastes into a deliverable. The note is now minted blank, linked, and only then given a body built from what actually landed. `DiffController#selected_record` hands back `{flow_id, builder}` rather than a finished draft so the text *can* be built after the fact, and the toast names a refused link instead of staying quiet about it.

**`create_note` refreshes `link_preview`.** It is one view-level string, not per note, so a new current note inherited the previous one's evidence line whenever nothing was linked — every `removed` row. `notes_duplicate` already called this for the same reason.

**`gone` no longer reads "every answer was 404, pending".** `Compare.status_label` appends `pending` for a flow that never got a response, and pending is the same "we have no evidence" state the removed/gone split exists to keep out of a deliverable. `b_statuses` now uses `sorted_statuses`, and the pending case gets its own clause. The verdict itself is `Facts#absent?`'s call and is out of this issue's scope, so it is untouched — only the claim the record makes about it.

**The A/B legend.** `Render.heading` says "B"; unlike the report it was written for, a record prints no A/B coverage table above it, so the reader had to guess. The header now reads `Retest diff — A: q1 (baseline) → B: q3 (newer)`.

**`Record.endpoint` calls `Render.endpoint_label`** instead of respelling it — the same argument that made `facts_line` public in the first commit, which I then broke one method over.

**The MCP `diff_projects` schema documents `observation`.** The field shipped without the contract mentioning it, which is the schema-vs-reader drift of #724 — the CHANGELOG sold it as what keeps an agent from filing a removal, while the only thing the agent reads never pointed at it.

**`body_hint` tracks availability.** All three row verbs gate on `diff_rows_shown?`, and a lens can empty the list. The hint now drops them, which also fixes `↵ Comparer` — already wrong before these two joined it.

**CHANGELOG cut to two sentences** per AGENTS.md ("If it has to be read twice, it is too long").

Also corrected the `stay_on_create?` doc: `link_ref` still decides its own path first, so the flag is read only when no ref was attached. (The reviewer's scenario — a caller setting both gets "the old jump-to-Issues behaviour" — is overstated: it gets the open-vs-stay confirm, which is also a stay. But the comment did overclaim the split, so it now states the actual contract.)

## Not taken

**`unchanged` should file at `info` like `removed`.** No. `removed` is `info` because a coverage gap is *not a finding* — that is the entire reason it is a separate verdict. An `unchanged` row filed as an issue is the opposite: "the thing I filed last quarter is still there", which is a confirmation. Medium is the form's default and the operator moves it.

**Per-row `observation` bloats the JSON** (~4.4 MB at MCP's 20k cap). The cost is real and the math is right, but every alternative offered trades away the property the field was added for. A report-level `verdict → sentence` map is reachable only by an agent that knows to look there — which is precisely how the report-level `caveats` already fails. Emitting it only on the absence verdicts means a consumer cannot read `row["observation"]` unconditionally. A 20k-endpoint row array is already megabytes of statuses, content types and ISO timestamps; this is a proportional add on a surface that is large by construction.

**`n` over a long sweep is O(n²) writes.** True, and it is the Notes model's shape, not this key's: the document is one settings row and `^N` / `create_blank_note_id` pay the same cost per write. At realistic retest volumes (tens of rows worth noting) it is tens of KB. At the volume where it would matter, 200 sub-tab chips are the problem well before the blob is.

**`require "../issues_export"` pulls the SARIF chain into the diff engine.** Fair, and `Render.iso` refuses a `CLI::` dependency twelve lines away for exactly this reason — but `src/gori/notes.cr` already requires it for the same three-line helper, so the precedent in the engine layer is established. Moving `scrub_controls`/`one_line` to a neutral home means touching `issues_export.cr` and its callers, which is a separate change.

## Verification after the pass

`crystal build --no-codegen src/main.cr` · `crystal spec` (**11,481 examples, 0 failures**) · `scripts/bench_check.sh` (48 harnesses) · `crystal build --no-codegen scripts/seed_demo.cr` · `crystal tool format` · ameba clean on every file I touched (the 3 left in `notes_controller.cr` are pre-existing complexity warnings on methods this branch does not touch).

Five new examples: the scrub chokepoint against a hostile `Content-Type`, the A/B legend, the `gone`-with-pending wording, `Record.endpoint == Render.endpoint_label`, and that the builder produces different evidence text for a link that landed vs one that did not — plus the hint dropping the row verbs on an empty lens.

Issue #844, opened 2026-08-28
feat(tls): choose a fingerprint per send, not only per destination


hahwul:
## Summary

Outbound TLS fingerprint control (#822) is keyed by **destination host pattern** and nothing else:

```
src/gori/settings/outbound_tls.cr    Settings.outbound_tls_for(dest_host)
src/gori/proxy/upstream.cr:744       client_context(verify, alpn, tls) # cached by {verify, alpn, tls.cache_key}
```

Neither `src/gori/repeater/` nor `src/gori/fuzz/` references it — they inherit whatever the destination rule says, through `Upstream`.

That granularity is right for the standing case ("always look like Chrome to this origin") and wrong for the question the feature exists to answer.

## The workflow it cannot express

The reason to have fingerprint control is to find out whether a target treats clients differently by handshake:

> Does this endpoint answer differently as `chrome` than as `curl`?

That is an **A/B on one host**. Today it requires editing a global destination rule between two sends, which means the two sends are not comparable (different settings state, no record of which was which), and it changes the fingerprint for every other tab and every background capture hitting that host at the same time.

The same shape appears in a fuzz run: "sweep this parameter with a browser-shaped hello" is currently expressible only by changing global state for the duration of the run.

## Design

Add a **per-send override** that names an existing preset or rule, resolved at send time, without touching the destination table.

- **Repeater**: a per-tab fingerprint selector, shown on the request card the way the other send modifiers are, and carried in the persisted repeater so a reopened tab sends what it sent before. Two tabs against the same host with different fingerprints is the case that must work — it is the whole point.
- **Fuzzer**: a run-level `--tls-preset` / config field, recorded on the run so a result set says which handshake produced it.
- **Comparer**: two sends that differ only by fingerprint should be diffable. This mostly falls out if the override is recorded on the exchange, but check it.

The override **narrows** the destination rule rather than replacing the whole policy: a client certificate configured for that host must still be presented. Decide and document what an override means for each field of `OutboundTlsRule` — a fingerprint preset legitimately overrides `groups`/`sigalgs`/`ciphersuites`/`alpn`, and legitimately does **not** override `client_cert`/`client_key`.

### The trap

[...]

PR #853, opened 2026-08-28
feat(miner,fuzz): process hooks reach the Miner, and the rest of the draw-vs-send contract (#846, #852)

### LINKED ISSUE #846: feat(miner,fuzz): process hooks do not reach the Miner, and only reach the Fuzzer by marker
## Summary

Process hooks (#818) reach three seams — the Rewriter's `pipe` op, the Decoder chain's `exec:` step, and Probe's `exec` rule. Two workbenches were left out, and the gap is uneven rather than principled:

| Surface | Reaches a hook? |
|---|---|
| Rewriter | ✅ `pipe` op |
| Decoder | ✅ `exec:` step |
| Probe | ✅ `exec` rule |
| Fuzzer — per-position `¦chain` | ✅ (via the Decoder registry) |
| Fuzzer — run-wide `--encode` / `processors` | ❌ |
| Miner | ❌ (no `ProcessHook` reference anywhere in `src/gori/miner/`) |

## The Fuzzer half is small

`src/gori/fuzz/plan.cr` already handles `exec:` properly — `refuse_unusable_chains` recognises an exec step, validates that its argv tokenizes, and deliberately does *not* ask the registry about it (the comment explains why: an exec step resolves to no converter by construction, and whether the command exists is a run-time answer). So a marker chain `§value¦exec:./sign.sh§` works today.

What has no hook is the **run-wide** pipeline: `Fuzz::Processor` is a closed catalog (`Prefix`, `Suffix`, `RegexReplace`, `Encode`, `Case`, `Hasher`) applied to every payload at every position via `--encode` / MCP `processors`.

The workaround — put `¦exec:…` on every marker — exists, which is why this is low priority. But it is worse the more positions a template has, and the asymmetry is the kind that reads as an oversight rather than a decision. Either add an exec processor to the catalog, or write down in `fuzz/payload.cr` why the marker chain is the only supported route.

## The Miner half is the real one

The Miner has no hook of any kind. That matters more than it looks, because hidden-parameter discovery is exactly where an engagement-specific transform is needed: an app that requires every parameter to carry an HMAC, a signed envelope, or a per-request nonce cannot be mined at all today — every candidate is rejected by the app before the Miner learns anything about it.

The Miner already injects candidates (`src/gori/miner/inject.cr`) and fingerprints responses. A hook at the injection point — "here is the request I am about to send, hand me back the one to actually send" — is the same contract `pipe` already has, and it is the difference between the Miner working and not working against signed APIs.

## Design

- **Miner**: a hook at the injection seam, taking the assembled request and returning the request to send. Same `Gori::ProcessHook` primitive, same P6 contract — a hard timeout, and a failure means the candidate is skipped with a notice rather than the whole mine wedging. Note the cost model: the Miner's own perf note says it is latency-bound and counts serial RTTs; a hook that forks per candidate changes that arithmetic, so measure and document where the cost lands (the same disclosure #818 made for Probe's `exec` running per flow on the analyzer fiber).
- **Fuzzer**: either an `exec` `Processor` in the catalog, or a comment in `fuzz/payload.cr` recording that the per-position chain is the supported route and why. Both are acceptable outcomes; pick one and argue it.

## Where

| Piece | File |
|---|---|
| Hook primitive (reuse) | `src/gori/process_hook.cr`, `src/gori/settings/hooks.cr` |
| Miner injection seam | `src/gori/miner/inject.cr`, `src/gori/miner/` engine |
| Miner config surfaces | `src/gori/tui/mine_config_overlay.cr`, `src/gori/cli/run/mine.cr`, `src/gori/mcp/tools/mine.cr` |
| Fuzz processor catalog | `src/gori/fuzz/payload.cr` |
| Existing exec handling (reference) | `src/gori/fuzz/plan.cr` (`refuse_unusable_chains`) |

## Design principles to honor

- **P6**: the hook's timeout is a budget with a bound the operator can predict. A mine over N candidates must not be able to multiply one timeout by N without saying so — #818 hit exactly this in the Rewriter and answered it with a per-message budget rather than a per-spawn limit. Decide the equivalent unit here and state it.
- **No shell**: argv exec'd directly, same as every other seam.
- **P1**: one primitive; the Miner does not grow its own process runner.
- **Absence of a finding reads as clean**: a hook that fails to spawn must not make a candidate look like a clean negative. #818 made this argument for Probe's `exec` rule and it applies identically here.

## Acceptance criteria

- [ ] A Miner run can transform each candidate request through an external command before it is sent.
- [ ] A hook failure skips the candidate with a reported reason; the run continues and the mine does not wedge.
- [ ] The timeout unit for a mine is documented and bounded — a slow hook cannot multiply into an unbounded run.
- [ ] The Fuzzer's run-wide pipeline either gains an exec processor or gains a comment recording why it does not.
- [ ] Where the hook cost is paid is documented, as #818 did for Probe.

## Scope

Small and mechanical relative to #818 — the primitive, the settings, and the failure semantics all already exist. This is reach, not new machinery.

---

### 한글 요약

#818의 프로세스 훅은 Rewriter(`pipe`)·Decoder(`exec:`)·Probe(`exec`) 세 이음매에 닿습니다. 워크벤치 둘이 빠졌는데 그 빠짐이 원칙적이지 않고 들쭉날쭉합니다.

**Fuzzer 쪽은 작습니다.** `fuzz/plan.cr`의 `refuse_unusable_chains`가 이미 exec 스텝을 제대로 처리해서(argv 토크나이즈만 사전 검증하고 명령 존재 여부는 런타임 답이라 레지스트리에 안 물음 — 주석에 이유까지 있음) `§value¦exec:./sign.sh§`는 오늘도 됩니다. 없는 건 런 전체에 걸리는 `--encode`/`processors` 파이프라인(`Prefix/Suffix/RegexReplace/Encode/Case/Hasher` 닫힌 카탈로그)뿐이고, 마커마다 붙이는 우회가 있어서 우선순위는 낮습니다. 다만 위치가 많을수록 나빠지고, 이 비대칭은 결정이라기보다 누락으로 읽힙니다 — 카탈로그에 exec 프로세서를 넣든지, `fuzz/payload.cr`에 "마커 체인이 유일한 경로인 이유"를 적든지 하나를 고를 것.

**Miner 쪽이 진짜입니다.** 훅이 아예 없는데, 숨은 파라미터 발견이야말로 engagement 고유 변환이 필요한 자리입니다 — 모든 파라미터가 HMAC·서명 봉투·per-request nonce를 달아야 하는 앱은 **오늘 전혀 마이닝이 안 됩니다**. 후보가 앱에 도달하기 전에 전부 거절돼서 Miner가 아무것도 못 배웁니다. 주입 지점(`miner/inject.cr`)에 "보내려는 요청을 줄 테니 실제로 보낼 요청을 돌려달라"는 훅을 다는 건 `pipe`와 똑같은 계약이고, 서명 API 상대로 Miner가 되느냐 마느냐를 가릅니다. **비용 모델 주의**: Miner는 지연 바운드에 직렬 RTT를 세는 구조라 후보마다 fork하면 그 산술이 바뀝니다 — #818이 Probe `exec`에 대해 했던 것처럼 비용이 어디서 발생하는지 문서화하고, **타임아웃 단위**도 정할 것(#818은 Rewriter에서 per-spawn이 아니라 메시지당 budget으로 답했고, 여기서도 N개 후보가 타임아웃을 N배로 곱하지 못하게 해야 합니다).

### LINKED ISSUE #852: fix(fuzz,repeater): the rest of the draw-vs-send hook contract — #851 follow-ups
## Summary

#851 fixed the four seams that forked the operator's command while merely **drawing** a chain, and its PR body listed four things it deliberately left for a follow-up. Three are inside `src/gori/fuzz/`, which another lane owned at the time. All four are verified still present on `main`.

Each is small. Together they are the rest of the "a hook runs when you SEND, not when you look" contract.

## 1. A refusal that lands after the side effect

The per-value chain refusal still runs **after** the hook has fired. Two markers on one template — one an `exec:` hook, one a pure chain that fails on its value — and the hook has already run by the time the pure chain's failure refuses the send.

The fix #851 named: **order pure chains first**, so a refusal that can be decided without a side effect is decided before any side effect happens. This is the same argument `Fuzz::Plan.refuse_unusable_chains` already makes about plan-time refusals ("a refusal before the first dial is the only report a sweep of ten thousand requests can act on"), applied one level down to the per-value pass.

Where: `src/gori/fuzz/template.cr` (`apply_chains_reported`, the class-method form).

## 2. Two validators that are verbatim twins

```
src/gori/tui/repeater_view/send.cr:184   refuse_unrunnable_chains(tmpl, registry)
src/gori/fuzz/plan.cr:691                self.refuse_unusable_chains(positions, registry)
```

`RepeaterView#refuse_unrunnable_chains` and `Fuzz::Plan.refuse_unusable_chains` ask the same question, in the same words, and produce the same envelope (`§…§ chain cannot run: …`). Two copies of a refusal sentence is exactly the drift `AltSvc`'s shared `removal_note` / `kept_note` exists to prevent.

**Mind the layering contract** (`DESIGN.md §2.1`): `tui/` may depend on `fuzz/`, never the reverse. The shared validator therefore belongs in `fuzz/`, with the TUI calling it — not a new module that both reach sideways into.

## 3. `Generator#baseline_raw` is not memoized

`src/gori/fuzz/generator.cr:118` runs `chained(@marked.default_payloads)` on every call, and `Plan.build` calls it to answer three separate structural questions (`Outbound.request_target`, `GrpcVerdict.framed_template?` / `reframable_template?`, `Plan#rewrites_content_length?`). With an `exec:` chain in the template that is **three forks of the operator's command to answer three questions about shape**, at plan time, before the run starts.

Memoize it. The method's own comment already frames it as "a bounded, plan-time cost" — it should be bounded at one.

## 4. `WsScript#apply_chains` does not take `run_hooks`

`src/gori/fuzz/ws_script.cr:109`. No caller needs it today, which is why #851 left it: a future caller fails to compile rather than silently forking. That is a reasonable holding position, but it is a trap primed to be defused by whoever hits the compile error under deadline. Add the parameter and forward it, the same way `GrpcFieldTemplate` now does.

## Design principles to honor

- **The contract**: a hook runs when you SEND, not when you look. Every one of these is that sentence.
- **Refusals precede side effects.** Item 1 is the general form of a rule gori already follows elsewhere; see the "refusal must precede side effect" reasoning in the intercept/fuzz paths.
- **§2.1 layering**: the shared validator lives in `fuzz/`, called by `tui/`.
- **One spelling per fact**: the refusal sentence has one author.

## Acceptance criteria

- [ ] A template with one hook marker and one failing pure chain refuses **without** having run the hook; a spec pins that the hook ran 0 times.
- [ ] `refuse_unrunnable_chains` and `refuse_unusable_chains` are one validator, living in `fuzz/`, with the TUI calling it.
- [ ] `Plan.build` forks a hook at most once for its structural questions; a spec pins the count.
- [ ] `WsScript#apply_chains` / `#apply_chains_reported` accept and forward `run_hooks`.
- [ ] Existing #851 regression specs still pass unchanged.

## Scope

No behaviour changes beyond the four items. Counting specs, not "few" — #851 established that only an execution count distinguishes these defects from correct behaviour, and every fix here should be pinned the same way.

---

### 한글 요약

#851이 체인을 **그릴 때** 운영자 명령이 fork되던 네 지점을 고치면서, PR 본문에 후속 4건을 남겼습니다. 셋이 `src/gori/fuzz/` 안이라 당시 다른 레인 소유여서 미뤄졌고, 네 건 모두 `main`에 그대로 있는 것을 확인했습니다.

1. **거부가 부작용 뒤에 옴** — 마커 둘(하나는 `exec:` 훅, 하나는 값에서 실패하는 순수 체인)이면 순수 체인의 실패로 전송이 거부될 때 훅은 이미 실행된 뒤입니다. #851이 지목한 수정은 **순수 체인을 먼저 정렬**해서, 부작용 없이 내릴 수 있는 판정을 부작용 전에 내리는 것. `refuse_unusable_chains`가 plan 시점에 대해 이미 하는 논증("만 건 스윕은 첫 dial 전에 거부돼야 한다")을 한 단계 아래 per-value 패스에 적용하는 것입니다.
2. **검증자 둘이 축자 쌍둥이** — `RepeaterView#refuse_unrunnable_chains`(send.cr:184)와 `Fuzz::Plan.refuse_unusable_chains`(plan.cr:691)가 같은 질문을 같은 문구로 하고 같은 봉투(`§…§ chain cannot run:`)를 만듭니다. **레이어링 주의**(§2.1): `tui/`는 `fuzz/`에 의존해도 되지만 반대는 안 되므로, 공유 검증자는 `fuzz/`에 두고 TUI가 호출.
3. **`Generator#baseline_raw` 메모이즈 없음**(generator.cr:118) — `Plan.build`가 구조 질문 3개에 이걸 3번 부르고, 템플릿에 `exec:` 체인이 있으면 **모양을 묻는 질문 3개에 운영자 명령을 3번 fork**합니다. 주석이 이미 "bounded, plan-time cost"라 부르고 있으니 1로 묶을 것.
4. **`WsScript#apply_chains`에 `run_hooks` 없음**(ws_script.cr:109) — 지금은 호출자가 없어 안전하지만, 마감에 쫓기는 사람이 컴파일 에러를 그냥 없애는 방식으로 풀 함정입니다. `GrpcFieldTemplate`이 이번에 한 것과 같은 방식으로 파라미터를 추가해 전달할 것.

전부 "훅은 **보낼 때** 돌지, **볼 때** 돌지 않는다"는 한 문장입니다. #851이 확립한 대로 **횟수를 고정하는 스펙**으로 검증할 것 — 이 결함들은 실행 횟수로만 올바른 동작과 구분됩니다.

hahwul:
Closes #846. Closes #852.

One lane, two issues, because they touch the same files. #852 (hygiene on code that already ships) landed first, so #846 adds a fifth `ProcessHook` caller to a seam that is actually true everywhere.

## #852 — the rest of the draw-vs-send hook contract

Four items #851 named and left for a follow-up, all verified present on `main`, each pinned by an **execution count** (`spec/fuzz/hook_draw_send_spec.cr`):

1. **A refusal that landed after the side effect.** `Template.apply_chains_reported` ran chains in position order, so a hook marker beside a pure marker that fails on its value fired the hook *before* the pure failure refused the send. Now two passes: everything decidable without a side effect runs first, and the deferred command positions do **not** run when a pure one already failed — a refusal decidable without a side effect is decided without one.
2. **Two validators that were verbatim twins.** `RepeaterView#refuse_unrunnable_chains` and `Fuzz::Plan.refuse_unusable_chains` asked the same question in the same words and raised the same envelope. There is one validator now, `Fuzz::Plan.refuse_unrunnable_chains`, living in `fuzz/` with the TUI calling it (**§2.1**: `tui/` depends on `fuzz/`, never the reverse), and one envelope author (`ChainError.unrunnable`) for both the template-level and per-value refusals.
3. **`Generator#baseline_raw` was not memoized.** `Plan.build` asks it three structural questions, so a template with an `exec:` chain forked the operator's command three times to ask about shape, at plan time. Memoized — bounded at one, as its own comment always claimed.
4. **`WsScript#apply_chains` did not take `run_hooks`.** Added and forwarded, the way `GrpcFieldTemplate` does — before the future caller that would have defused the compile error by silently forking.

## #846 — reach

**The Miner half is the real one.** Hidden-parameter discovery is exactly where an engagement-specific transform is needed: an app that requires every parameter to carry an HMAC, a signed envelope or a per-request nonce rejects every raw candidate before reacting, so today it cannot be mined at all. `gori run mine --hook './sign.sh'` and the MCP `mine_start` `hook` argument pipe each assembled request through the operator's own command before it ships.

Wired as a `Fuzz::Backend` **decorator** (`Miner::HookBackend`), not a call at one inject site, because a mine's sends do not all leave from one place — baseline calibration, bucket probes and confirmation all go through `@backend.send`, and the baseline **must** be hooked too (an unsigned calibration probe against a signed API is "unreachable"). Order is **bindings, then hook, then byte-exact**: session tokens resolve first (injected-candidate spans keep candidates literal), so a signature covers the real wire bytes, and the hook's output ships as `all_verbatim` so nothing below rewrites it.

- **Failure is a skip with a reason, never a clean negative.** A hook that cannot spawn / times out / exits non-zero comes back as an errored send, which every miner send site already refuses to read as a confirmed absence (#818's "absence of a finding reads as clean"). The run continues; it does not wedge. A bad argv is refused at `Plan.build`, not per-worker.
- **P6 — the timeout unit is per outbound request**, the shared `hooks.timeout_secs` budget. A mine over N candidates cannot multiply one timeout by N unseen: the number of hook runs *is* the number of requests, which the miner already counts (`Progress#sent`) and bounds (`--max-requests`, the finite bucket/bisection/confirm tree). `HookBackend` sits inside `CappedBackend`, so once the budget is spent the cap refuses before the hook forks.
- **Where the cost lands.** The miner is latency-bound; a hook adds one serial fork+exec+wait to each RTT — documented, the same disclosure #818 made for Probe's `exec`.
- **TUI parity.** The mine config overlay is deliberately field-free (no IME plumbing), so `--hook` (and `--wordlist`) are a known CLI/MCP-only gap — recorded in `build_config`.

**The Fuzzer half ends in a comment.** `§value¦exec:./sign.sh§` already works. The run-wide `--encode`/`processors` catalog stays pure-transform on purpose, argued in `fuzz/payload.cr`: a `Processor` is a `String -> String` with no failure channel, so an exec there could only raise (kill the run) or pass through silently (the clean-negative bug); the per-position chain already has the honest channel (`Job#chain_error`), the `run_hooks` withholding, and the up-front argv validation a closed catalog cannot.

## Shared file with the #844 lane

Item 2 put me in `src/gori/tui/repeater_view/send.cr`: I **deleted** the private `refuse_unrunnable_chains` twin (now `Fuzz::Plan.refuse_unrunnable_chains`) and changed `refuse_chains` to raise the shared `ChainError.unrunnable`. No migration taken (`V22` is the #844 lane's).

## Verification

- `crystal build --no-codegen src/main.cr` ✅
- `crystal spec` — 11600 examples, 0 failures ✅
- `./scripts/bench_check.sh` — 48 harnesses ✅
- `crystal build --no-codegen scripts/seed_demo.cr` ✅
- `crystal tool format --check` clean; `ameba` adds no new findings on the changed files (the two remaining `not_nil!`/`else nil` are pre-existing lines I did not touch)
- New counting specs fail on the pre-fix tree the way #851's do; the #851 regression specs pass unchanged.


hahwul:
Addressed the `/code-review` findings in c8d1c84e:

- **Slot overlay ordering (the real one).** Confirmed and fixed. `Fuzz::Sender`'s `@slot_overlay` **defaults to `true`**, so the miner's sender *was* applying the session-slot header overlay *after* the hook — a `--slot analyst --hook ./sign.sh` run against a signed API would have shipped an overlay the signature couldn't cover, and the miner would read the rejections as a clean negative. `HookBackend` now applies the overlay itself **before** the hook, and `Plan.build` builds the inner sender with `slot_overlay: false` when a hook is present so it's applied exactly once. Pinned by a spec whose hook rewrites the overlaid header (seeing the rewrite proves the overlay ran first) plus a wiring assertion. Good catch — this is the class of thing the "review round on #818 caught an RCE" note is about.
- **Hook forks before the scope gate.** Documented honestly in the class comment rather than restructured: the Sandbox/scope gate lives in the sender with the blocked-count accounting it owns, and a wholly out-of-scope origin (the miner sends to one fixed origin, so scope is all-or-nothing) fails **calibration** and is refused before any candidate is tested — so the forks are bounded to calibration, not one per candidate. The cap, which *can* multiply per candidate, is outside `HookBackend` and refuses before the hook forks.
- **Stale `refuse_unusable_chains` references.** All 10 renamed to `refuse_unrunnable_chains`.
- **Doubled `parse_spec` in the two-pass.** Left as-is: avoiding it needs a tokens-accepting `Decoder.run` overload (a core-method refactor), disproportionate to ~2 small allocations per send on a network-bound, chained-only path — and the repo's own guidance is that CPU micro-opts don't pay.

`crystal spec` (11602, 0 failures), `format --check`, and `ameba` on the changed files are all green.

Issue #822, opened 2026-08-27
feat(tls): outbound TLS fingerprint control (JA3/JA4) and browser presets


hahwul:
## Summary

When an operator puts gori in the loop, the ClientHello the origin sees is **gori's OpenSSL handshake, not the browser's**. Anti-bot stacks (Cloudflare, Akamai, DataDome, PerimeterX) fingerprint that handshake (JA3 / JA4) and start serving challenges or 403s to traffic that worked fine without the proxy. The result is the everyday frustration of "it works in the browser, it breaks through the proxy" — and the operator has no knob to reach for.

**Neither Burp nor Caido solves this**, and it is hard to solve with extensions in either. It is a genuine differentiator, and gori is unusually close to it: `Settings::OutboundTlsRule` is already a per-destination TLS policy table, `apply_outbound_tls` is already the single choke point, and the repo already carries a real `LibSSL`/`LibCrypto` FFI layer (`src/gori/proxy/tls/ffi.cr`).

## Motivation

- Today's knobs (`src/gori/settings/outbound_tls.cr`): `client_cert`, `client_key`, `min_version`, `max_version`, `ciphers`, `permissive`. Enough to reach a legacy appliance; nowhere near enough to look like Chrome.
- The operator's current workaround is to stop using the proxy — which is to say, to stop using gori.

## Design

Staged, because a byte-exact uTLS-grade match is not reachable from stock OpenSSL and pretending otherwise would ship a knob that silently does not work.

### Phase 1 — knobs + observability (do this first)

Extend `OutboundTlsRule` with the fields OpenSSL *can* honor, applied in `Upstream.apply_outbound_tls` (`src/gori/proxy/upstream.cr`):

- `groups` / curves order — `SSL_CTX_set1_groups_list` (e.g. `X25519:P-256:P-384`)
- `sigalgs` — `SSL_CTX_set1_sigalgs_list`
- `ciphersuites` — TLS 1.3 suites (`SSL_CTX_set_ciphersuites`; note the existing `ciphers` field is TLS ≤1.2 only, as that file's comment already documents)
- `alpn` list + order — today only a single protocol is offered (`ctx.alpn_protocol = alpn`); a browser offers `h2,http/1.1`
- `session_tickets`, `ocsp_stapling`, `sni` override toggles

Most of these need FFI beyond Crystal's stdlib — add them to `src/gori/proxy/tls/ffi.cr`, which already exists for exactly this reason.

[...]

PR #854, opened 2026-08-28
feat(tls): choose a fingerprint per send, not only per destination (#844)

### LINKED ISSUE #844: feat(tls): choose a fingerprint per send, not only per destination

hahwul:
Closes #844.

## The problem

#822 gave gori per-destination TLS fingerprint control. That granularity is right for a standing policy — *"always look like Chrome to this origin"* — and wrong for the question the feature exists to answer:

> Does this endpoint answer differently as `chrome` than as `curl`?

That is an **A/B on one host**. Today it needs a global destination rule edited between two sends, which means the two sends are not comparable (different settings state, no record of which was which) and it changes the fingerprint for **every other tab and every background capture** hitting that host at the same time.

Neither `src/gori/repeater/` nor `src/gori/fuzz/` referenced `outbound_tls` at all — they inherited whatever the destination rule said, through `Upstream`.

## What lands

A **per-send override**: a Repeater tab and a fuzz run can each name a fingerprint, resolved at dial time, without touching the destination table.

| Surface | How |
|---|---|
| Repeater tab | `␣T` cycles none → chrome → firefox → safari → curl → none. A `␣T:…` chip rides the TARGET band beside `^V` (clickable). Persisted in `repeaters.tls_preset` (**Schema V22**) so a reopened tab sends what it sent before. |
| `gori run repeater` | `--tls-preset NAME` on the flow replay, `send`, `create` (stored on the session) and `h2`. `--format json` reports `tls_preset`. |
| `gori run fuzz` | `--tls-preset NAME`, run-level. The run banner names it: `fuzzing https://… · 300 requests · sniper · tls chrome`. |
| TUI Fuzzer | a **TLS fingerprint** row on the `^O` advanced card, persisted in the session's config JSON. |
| MCP | `send_request{tls_preset}`, `send_websocket{tls_preset}`, `fuzz_start{tls_preset}`, `create_repeater`/`update_repeater{tls_preset}`. `fuzz_start` / `fuzz_status` / `send_request` echo it back, so a result set says which handshake produced it. |
| Reporting | `gori settings tls-fingerprint HOST --preset NAME` prints the JA3/JA4 an override actually sends, narrowing each reported policy exactly the way a send does. |

Two tabs against the same host with different fingerprints is the case that has to work, and it is the whole point.

## The narrowing rule, field by field

The override **narrows** the destination policy rather than replacing it:

| Field | Under an override |
|---|---|
| `preset`, `groups`, `sigalgs`, `ciphers`, `ciphersuites`, `alpn`, `session_tickets`, `ocsp_stapling` | **replaced** by the named preset's |
| `client_cert`, `client_key` | **kept** |
| `min_version`, `max_version` | **kept** |
| `permissive` | **kept** |

The certificate is the one the issue calls out and the reasoning generalises: an override says what the hello should *look like*, not *who gori is* — dropping the client certificate would turn "chrome vs curl" into "authenticated vs anonymous" and answer a different question than the one asked. The version range is a reachability fact about that destination (a TLS 1.0-only appliance stays reachable under any preset), and `permissive` is the one knob that *widens* what gori accepts, so an override must be able neither to grant it nor to revoke it.

**Replaced, not merged**, for the shape half. A rule's own fields beat its preset (`effective_groups` and friends), so a merge would leave the destination's `groups` winning and the override could not change the one thing it exists to change. `ciphers` goes with the shape because it is the TLS 1.2 cipher list *and its order*, a primary JA3 input.

## The trap

`client_context` memoizes SSL contexts by `{verify, alpn, tls.cache_key}`. #822's own review named that the most likely bug in the issue, and an override reintroduces it from a new direction: **if the override does not participate in the cache key, two sends with different fingerprints silently share one SSL context** — and the operator compares two responses that came out of one handshake, which is exactly the experiment the feature exists to make possible.

It is closed by construction rather than by being remembered. The override reaches the dial as a *narrowing of the same `OutboundTlsRule` every dial already resolves*:

```crystal
tls_policy = Settings.outbound_tls_for(host, tls_preset)
```

so `cache_key` covers it because `cache_key` covers the record. There is no second TLS policy path (P1) — `apply_outbound_tls`, `cache_key` and the fingerprint report all keep reading exactly one record and cannot tell where its shape came from.

`spec/proxy/outbound_tls_override_spec.cr` (15 examples) proves the behavioural half, not just the key:

- two different overrides on **one host** get two contexts whose JA3 *and* JA4_r differ, and the same override twice is still one context;
- an overridden send does not share the destination rule's context, while an override naming what the rule already says correctly *does* (the key separates policies, not requests);
- every preset gets its own context, pairwise — a key that folded two of four together would pass a single-pair test;
- two destinations' client certificates stay apart under one override;
- **no override at all is byte-identical**, same `cache_key`, same cached context.

## Verified live

Against the real `example.com`, with a standing `preset: "firefox"` rule on that host, so the override is proved to beat a *configured* policy and not just to beat "nothing configured":

```
(none — the destination rule)  ctx=0x1049dec90 ja3=0410655d3476b38ad944e58c44dee1ae ja4=t13d1713h2_5b57614c22b0_e5d904db1611
                               handshake: TLSv1.3 alpn=h2
chrome                         ctx=0x1049acf00 ja3=c99e92e692ba483e2602b38b3c0a5645 ja4=t13d1513h2_8daaf6152771_afafd945c4ab
                               handshake: TLSv1.3 alpn=h2
curl                           ctx=0x1049acdb0 ja3=b917dec3287a9a5b2bde86bf73b3ef1d ja4=t13d3011h2_1d37bd780c83_0e755cee3882
                               handshake: TLSv1.3 alpn=h2
safari                         ctx=0x1049acd80 ja3=8777ffa199aefa6ecd84912f15a3c89f ja4=t13d1713h2_5b57614c22b0_afafd945c4ab
                               handshake: TLSv1.3 alpn=h2
```

Four distinct contexts, four distinct JA3, four real TLS 1.3 handshakes with verification on and h2 negotiated. The `chrome` override comes out as `c99e92e692ba483e2602b38b3c0a5645` — the same JA3 a `preset: "chrome"` **destination rule** produces, so an override and a rule naming the same preset really are one policy.

## Honesty and P4

An unknown preset name is **refused at every surface before anything dials** — `Plan.build` on both the Repeater and the Fuzzer side raises a machine-readable `TlsPreset` reason each surface words for itself, `repeater create` aborts, MCP returns `INVALID_ARGUMENT`, and `settings tls-fingerprint --preset` refuses too. This is the one case where silence is worse than a wrong value, for the reason #822 already argued about `parse_tls_preset`: an unknown name applies *nothing*, so the send goes out with gori's bare OpenSSL hello while the surface names a browser — and unlike the destination table, a per-send override has no startup warning to catch it. The name is kept verbatim rather than folded to `""`, so the refusal can name it.

**P4**: nothing infers a fingerprint. A tab or a run has one only because the operator set it, and the TUI cycles through known names so an unknown one can never originate there. An override left on a tab retargeted to `http://` is **kept** and reported as inert (a muted chip, and a status line saying so) rather than silently discarded — the operator set it, and retargeting back to `https://` must bring it back.

**The claim is not upgraded.** Every help string, schema description and doc paragraph repeats #822's: these are *approximations*, extension order and GREASE placement are OpenSSL's, and `gori settings tls-fingerprint HOST --preset NAME` is how you check rather than believe.

## Review follow-ups folded in

A pass over the first draft found three worth naming:

- **`update_repeater`'s SQL writes every column unconditionally**, and its signature defaults them — so the three callers that omit one CLEAR it. `tls_preset` would have joined `ws_keep_key`/`ws_http_only` as a fourth thing a plain rename or a `minimize --apply` silently drops. Threaded through MCP `update_repeater`, MCP `minimize_repeater` and `gori run repeater minimize`; the neighbouring comments already warned about exactly this and now name the third column.
- The chip's label consulted `parse_target`, putting a `URI.parse` on **every frame** for a question only the chip's *colour* needs. The label no longer looks at the scheme; `tls_preset_live?` does, and only when an override is actually set.
- `outbound_tls_for` had its own `strip.downcase` beside `tls_preset_normalize`. Two copies of "which spellings are one policy" is two answers to the cache-key question — routed through the one function.

## Not in scope

New preset definitions, or improving the fidelity of the existing ones — this is about *where* a fingerprint can be chosen, not *which* fingerprints exist. No `fuzz_runs` column either: that table has no production writer at all, so a column there would record nothing; a run reports its fingerprint through the same summary / JSON / MCP-job metadata that reports the rest of its settings.

## Verification

`crystal spec` **11622 / 0 failures** · `crystal build --no-codegen src/main.cr` · `./scripts/bench_check.sh` (48 harnesses) · `crystal build --no-codegen scripts/seed_demo.cr` · `crystal tool format --check` · ameba (no new findings on the changed files) · the live handshake run above.

Migration **V22** (`repeaters.tls_preset`), the number reserved for this lane; `main` ended at V21.


hahwul:
## Review round — all twelve findings addressed

`/code-review` at **xhigh** found twelve. Nine were behavioural, and the theme is one this codebase keeps rediscovering: **the override reached one path and not its twin.** This PR's own diff argued about that hazard twice (the `update_repeater` write, the ALPN legs) and left three more instances open.

### Behavioural (9)

| | What was wrong |
|---|---|
| **`send_request{tls_preset: ""}` could not clear** | `.presence \|\| stored` folded `""` to nil, so the fallback fired and the **baseline half of an agent's A/B went out as `chrome` too** — two identical handshakes, both reported as `chrome`, and an agent concluding the origin is fingerprint-insensitive from an experiment that never happened. The schema promised `""` would drop it. Now `present?`; `send_websocket` and `fuzz_start` got the same escape. |
| **All three Minimize surfaces sent with the wrong hello** | TUI, MCP and CLI each build a `Fuzz::Sender` from the session (sni, http2, evidence) and none passed the fingerprint — so up to `SEND_CAP` probes go out bare against the origin that motivated setting a preset. Every candidate gets the same WAF answer, the bisection reads that as *"every header is removable"*, and `--apply` rewrites the stored request from responses no real send produced. This PR had already threaded the column through those exact three functions on the **write** side. |
| **A fuzz run seeded from a tab dropped its fingerprint** | `fuzz_start{repeater_id}` and `gori run fuzz --repeater N` both inherit `rec.sni` — with a comment arguing a pinned session must be swept against *that* name — and inherited nothing for the handshake. Now both do, and an explicit `--tls-preset ""` still takes the baseline. |
| **A duplicated tab's row said "no override"** | `persist_new_repeater` omitted it, so until some later save-on-leave a peer session, `repeater list` and MCP all read nil while the chip on screen read `␣T:chrome`. |
| **The reconcile poll never converged on a non-canonical spelling** | `request_side_matches?` compared the view's *normalised* value against a raw row value, so a row holding `"Chrome "` was "changed" on every poll → `apply_peer_request` forever → the caret slammed each time. Precisely the failure the method's own new comment claimed to prevent. |
| **`gori run repeater --format json` named a hello it never sent** | No `https` guard, where MCP (`plan.scheme == "https"`) and the fuzz banner both have one and the docs promise *"gori will not report one it did not send"*. |
| **`repeater list` showed no fingerprint** | Neither text nor JSON — while `repeater create`'s refusal of an unknown name argues at length that a stored preset *is* visible there. Now appended to the text line only when set, so an override-free workbench prints the line it always did. |
| **`ws_keep_key` cleared by the TUI's own save** *(pre-existing)* | `save_current_repeater`'s non-WS branch and `persist_repeater_tab` omit it, and `update_repeater` writes every column unconditionally with a `false` default — so `repeater create --ws-keep-key -f plain.txt` loses the flag on the next TUI edit. Found while threading the *third* column through the same call. (`ws_http_only` is genuinely false there — the view forces it false for a non-upgrade request — so it is passed for symmetry, not to preserve anything.) |
| **`persist_send_repeater` stores unguarded while the reply guards** | Kept as-is, and the asymmetry is now stated in the code: the reply says what **this send** did (a plaintext send made no ClientHello), the row says what this **tab** is set to — the operator's choice, which survives a retarget to `https://` and is exactly what the muted chip reports (P4). |

### Hygiene (3)

- `tls_preset_live?` ran `Env.expand` + `URI.parse` on **every rendered frame** to pick one chip's colour. The chip's *label* had been deliberately kept off that path; the colour lookup put it straight back. Now a prefix test, which fails **safe** (chip goes muted) on the one target shape the two answers differ on — `$SCHEME://host`.
- `tls_preset_error` re-implemented `strip.downcase` rather than calling `tls_preset_normalize` — the duplicated decision `outbound_tls_for` argues against ten lines above it, and the one that would make validation and the cache key disagree about which names are valid.
- The Fuzzer's `Reason::TlsPreset` branch was annotated *"unreachable, the pane only cycles known presets"*. That row is a **free-text field** (`{:tls_preset, "TLS fingerprint", :text}`), so it is the ordinary path — an operator who types `chromee` hits it on the first run. Corrected rather than left to invite a future deletion.

### Specs

Three added, all pinning behaviour rather than shape: a non-canonical stored spelling converging on the reconcile poll, the scheme test across six target shapes (including `$BASE/path`), and the minimize backend carrying the fingerprint into both the pooled and unpooled sender.

`crystal spec` **11625 / 0 failures** · `--no-codegen` build · `bench_check.sh` (48) · `seed_demo.cr` · `crystal tool format --check`.

PR #834, opened 2026-08-27
feat(history): syntax-highlight request and response previews

sebastianosrt:
## Summary

- syntax-highlight HTTP request and response structure plus structured bodies in the History preview
- lazily memoize body highlighting in a bounded cache to keep preview rendering responsive
- invalidate cached highlight colours when the active TUI theme changes
- qualify the Repeater line-diff module in its side-by-side spec to avoid the new `Gori::Diff` namespace

## Behavior change

The History Request/Response preview now uses the same syntax colouring as the full detail view. Preview content, scrolling, and focus behavior are unchanged.

## Testing

- `crystal build --no-codegen src/main.cr` ✅
- `crystal tool format --check src spec bench scripts` ✅
- `crystal spec` ✅ — 11,210 examples, 0 failures
- `./scripts/bench_check.sh` ✅ — all 48 harnesses build

PR #855, opened 2026-08-28
fix(tui): decode compressed response views (#834)

sebastianosrt:
## Summary

- decode compressed response bodies in the History split preview and Fuzzer result detail through the existing `Entity` display seam
- keep the History preview bounded after decompression and align the Fuzzer detail ceiling with its decoded metrics
- preserve captured wire bytes and tolerant fallback behavior, with regressions for gzip, decoded-output truncation, malformed streams, and evidence retention

## Why

The History list preview and Fuzzer result detail were the two read-only TUI response projections still rendering retained wire bodies directly. A normal response carrying `Content-Encoding: gzip` therefore appeared as binary garbage even though History detail, Repeater, Comparer, and Authorize already displayed the decoded entity.

This change routes those missing views through the existing display-time decoder. It does not rewrite the stored response, forwarding path, replay bytes, headers, raw copy/export output, or hex views.

## Intentional behavior change

Read-only response text in these two panes now shows the decoded entity for chunked/gzip/deflate/Brotli/zstd responses. Unsupported or malformed encodings continue to fall back to the captured bytes instead of raising or blanking the pane.

## Validation

- `crystal build --no-codegen src/main.cr`
- `crystal tool format --check src spec`
- `crystal spec` — 11,460 examples, 0 failures, 0 errors
- `./scripts/bench_check.sh` — all 48 harnesses build


hahwul (approved):
LGTM

hahwul:
Hi @sebastianosrt 
Thanks for contributing and Great work! Merged

PR #858, opened 2026-08-28
feat(proxy,tui): configure socks proxy with authentication, route egress through upstream proxy

sebastianosrt:
## Summary

- Route updater and OAST HTTP traffic through the same upstream decision and dial path as the proxy/send engines, with no direct retry after a configured route fails.
- Distinguish `socks5` local DNS from `socks5h` proxy DNS across scalar settings, ordered rules, and both settings editors.
- Add project-scoped proxy authentication and a destination-host gate, keeping credentials attached to an explicit project proxy pin and failing closed on malformed/orphaned state.
- Add egress seam coverage so new direct socket or `HTTP::Client` construction cannot bypass the reviewed routing owners.

## Behavior changes

- Existing `upstream_rules` entries using `kind: "socks5"` now resolve destination names locally. Use `socks5h` to preserve the previous proxy-side DNS behavior.
- Malformed proxy declarations and failed HTTP CONNECT/SOCKS handshakes no longer degrade to a direct origin connection.
- A project destination non-match is explicitly direct and does not fall through to global rules or the scalar proxy.
- Project proxy passwords are stored in the owner-only project database and are masked in the TUI except while the password row is focused.
- Updater/OAST traffic follows the effective upstream route while continuing to skip target host overrides and the active-testing scope gate.

## Why

`network.upstream_proxy` looked like a catch-all, but updater and OAST stdlib clients could still open direct sockets. The new `HttpTransport` keeps those service clients on `Proxy::Upstream`, while `Settings.upstream_route(host)` remains the single place that chooses direct, HTTP CONNECT, SOCKS5, SOCKS5H, or an invalid fail-closed route. Project credentials pin the proxy address they were entered for so later global/rule edits cannot redirect the secret to a different proxy.

## Validation

- `crystal build --no-codegen src/main.cr` — passed
- `crystal tool format --check src spec` — passed
- `crystal spec` — 11,673 examples passed
- `./scripts/bench_check.sh` — all 48 harnesses build

Refs #434.

### DIFF
--- a/.github/DESIGN.md
+++ b/.github/DESIGN.md
@@ -1815,6 +1815,40 @@ lensed and an all-hidden list reads `no flows in scope` (plus `⇧S toggles the 
 For the same reason the open-site passes `raise_on_error: true`: `Store#search` otherwise degrades
 a SQLite failure to an empty result, which on this card is indistinguishable from an empty scope.
 
+### 2026-08-26: upstream proxying is fail-closed for app-owned egress
+
+Refines: [P1](#p1), [P4](#p4), [P5](#p5). #434.
+
+`network.upstream_proxy` used to govern the proxy/send engines but not the updater or OAST's
+stdlib HTTP clients. That made the setting read as a catch-all while two owned request paths
+could still leave directly. The routing decision now stays in `Settings.upstream_route(host)`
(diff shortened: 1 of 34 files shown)


hahwul (approved):

hahwul:
@sebastianosrt 
Nice work! Thanks for contributing :)

PR #860, opened 2026-08-28
feat(tui): search help and hotkeys panels

sebastianosrt:
## Summary

- Add explicit `/` search to the Help tab’s Shortcuts and Query pages, the Help reference popups, and the Hotkeys editor.
- Keep matching rows grouped under their section headings, show clear empty-result states, and share the Help row projection between the full tab and popup cards.
- Preserve each panel’s navigation state: Escape clears search and restores the prior scroll or selection, while Hotkeys Enter jumps back to the full editor at the selected binding.
- Route IME preedit, cursor ownership, bracketed paste, and command chords through the new search modes without changing raw hotkey capture.

## Behavior changes

- Help reference popups use explicit search: press `/` to enter search, Escape to clear it, and Escape again to close the popup.
- The Help tab searches only the visible Shortcuts or Query reference page and cancels search when switching pages.
- Help searches use case-insensitive substring matching across each row’s key/expression and description while retaining the matching section heading.
- Hotkeys search matches the visible scope, action title, and effective binding. Enter leaves search at the selected result without saving, so the binding can be edited before the overlay is committed.
- Hotkey capture still takes raw-key precedence, including binding `/`, and Ctrl/Alt chords continue to bypass text insertion.

## Why

The Help and Hotkeys references are long enough that navigation alone makes known entries hard to recover. An explicit `/` mode gives all three surfaces the same entry and cancellation model while keeping browse behavior intact. `HelpView.search_rows` is the shared projection for the full Help tab and popup cards, so their matching and section retention cannot drift.

## Validation

- `just build` — passed
- `crystal tool format --check src spec bench scripts` — passed
- `just test` — 11,695 examples passed
- `just benchmark-check` — all 48 harnesses build
- `just check` — format passed; Ameba reports its existing 906-finding backlog, with no findings in the changed source files


hahwul (approved):
LGTM

hahwul:
@sebastianosrt 
Great work! Thanks a lot!

Issue #864, opened 2026-08-29
feat(tui): Activity tab — a human window over the #124 event feed (agent audit + silent failures)


hahwul:
## Summary

The `events` table (#124) is the AI firehose — an append-only log of job lifecycle and **every agent mutation/send**. `log_agent_action` (`mcp/tools.cr:833`) records each agent tool call, success and failure, with its own comment declaring the purpose: *"record a completed agent mutation/send into the store event feed so the AI's activity is **visible to the human** (and tailable via list_events)."*

Except it isn't visible to the human. The only reader on that half of the story is `events_after` → the MCP `list_events` tool — an **agent-facing** forward cursor. `grep -rn "EventLog\|events_after" src/gori/tui/` returns **nothing**. The human side has the Agents overlay (who is attached) but no view of **what any of them did**. The feed's own stated reason for existing is unfulfilled on the surface it names.

## What's invisible today

Two classes of row are written to `events` and reach no human surface:

**Agent audit trail** — `mcp/tools.cr:839` logs every mutating/sending tool call (`create_issue`, `send_request`, `fuzz_start`, rule edits, …) plus scope-blocked/errored attempts. This is exactly the P4 audit record ("the human decides / can see what was decided on their project"), and right now the human can't read it in the TUI.

**Silent failures that never became a notification** — several `insert_event` sites write to the feed *only*, never to the notification ring:

- bindings `extract_miss` / `extract_scrubbed` / `extract_no_body` (`bindings.cr:1015,1026,1036`)
- bindings `unbound` / `boundary_refused` (`rules.cr:1217,1226`)
- Rewriter/Probe hook failures (`rules.cr:993`, `probe/custom_rule.cr:269`)
- Probe `alt_svc_h3` H3-bypass notice (`probe/analyzer.cr:580`)

A session-handling extraction that silently missed, a hook that failed and passed bytes through unchanged — the operator has no way to see any of it. The data is captured; the window is missing.

## Design — an Activity view over the existing feed

No new engine. One new read + one surface.

- **Read.** Add `events_recent(limit, before_id?)` to `store/reads.cr` — newest-first, paginated, the mirror of the existing oldest-first `events_after` (the same shape `recent_agent_actions` at `reads.cr:122` already uses, minus the `agent_action`-only filter). This is a **pull** surface (P8): a query over the log, not a queue that pushes.
[...]

PR #875, opened 2026-08-29
fix(tui,store): Activity pane holds its place on a large, live feed

hahwul:
Tested #864's Activity pane against a 200k-row event feed. Paging itself was sound — no duplicates, no gaps, page one in ~4 ms, `⇧C` clearing 200k rows in under half a second — but four things went wrong once the feed outgrew one screen or anything was writing to it.

### The walked-back scan was rewound every few seconds

A narrowing that matches nothing in the newest window is walked back one scan window per `↓`, and `refresh_activity` reloaded page one whenever the loaded list was empty — throwing the walk away. That is not a rare path: the capture-lock holder rewrites the intercept-bridge heartbeat every 3 s, so `data_version` moves on a project nobody is touching.

The two matching rows at the bottom of a 200k feed were reachable only by pressing `↓` faster than the heartbeat, and the pane insisted, permanently, that it had looked no further than the newest 50k events — the exact claim `activity_no_match_line` exists to keep honest.

```
before, 1.6 s between presses:  50k → 100k → 150k → 50k → 100k → 50k   (never arrives)
after,  same cadence:           50k → 100k → 150k → 200k → the two rows
after,  8 s idle:               200k held   (before: back to 50k within 3 s)
```

An empty list is now told apart from an unloaded one by `@act_walked`, because `next_before` cannot: it is nil both for a feed whose bottom the scan reached and for a feed that holds nothing at all. The rule is that a refresh may never move the cursor toward newer ids. New matches are still picked up at the head, which is the only place `id` being AUTOINCREMENT lets them appear.

### The pane did not tail, and did not say it wasn't

The id anchor held the cursor on its event even when that event was the head, so an operator watching the newest row saw a frozen screen while hundreds of events piled up above it, with only the border count moving.

A cursor **on** the head now follows it — "the newest row" is a position the operator chose, not the neighbour-slide the anchor exists to prevent — and one parked further down keeps its event and gets told, via `↑N new` on the card border. That number is capped at the rows actually above the cursor, so climbing through them empties it out instead of promising five new rows to someone standing on the second one.

### A burst larger than one page discarded the loaded pages

"More than 200 arrived at once" was read as "the list was truncated", so one 500-event burst snapped a list paged 1 250 rows deep back to 200 and dropped the cursor, via `clamp_act_sel`, on a row nobody had selected — during exactly the burst this pane exists to show. `refresh_activity` now walks down to `ACT_CATCHUP_PAGES` (5) pages looking for the head before it gives up, which covers an agent's job without letting the render fiber chase one forever.

### `EventPage#window` counted the scan bound, not the ids read

A page that fills at the LIMIT stops at its last row, so crediting it with the whole window overstated it by two orders of magnitude — and `activity_load_more` accumulates it, so walking a 200k feed reported 43.7M ids read for a sentence whose whole job is to name what the pane looked at. The sum over a full walk is now the size of the feed (200,000 for every filter tried).

### And the `⇧C` confirm was truncated at "…what each attac…"

`ConfirmDialog` splits on `\n` and caps the card at 60 columns, so the one dialog that has to finish naming what it destroys did not. Hand-wrapped like every other confirm in the app; ENABLE SANDBOX had the same break and is fixed with it.

## Verification

- `crystal spec` — 11800 examples, 0 failures (on the rebased tree, so the merge result itself is covered)
- `crystal tool format --check src spec`, `crystal build --no-codegen src/main.cr`, `scripts/bench_check.sh` (48 harnesses), `scripts/seed_demo.cr` — all clean; no new ameba findings
- 8 specs added: `window` accounting in `events_recent_spec`, and a describe block in `project_activity_pane_spec` covering the poll path — no rewinding, new head matches after a walk, a resume point for a feed empty at open, head following, the parked cursor's `↑N new` and its shrink-on-scroll, burst catch-up, and the fallback past the bound
- Re-verified in a real TUI over tmux against the 200k feed for each of the five behaviours above

Behaviour changes are folded into #864's existing CHANGELOG entry rather than added as a separate line, since that feature has not shipped yet.


hahwul:
CI shard `tests (1/4)` was red on the first push. Not a flake and not the pane change — a pre-existing cross-spec leak this PR happened to expose.

`spec/store/config_events_spec.cr`'s `cfg_store` restores `FlowSource.surface` and deletes its temp db, but a **global** Match & Replace rule lives in neither: `Settings` is a process-wide singleton, and `"records a scope move as a move"` calls `set_scope(…, RuleScope::Global)`, which moves the rule there. The leftover outlived the temp db, so `commit_confirmation_spec`'s `"refuses an empty pattern without claiming a write"` found a rule named `movable` in a `Rules` it had just built and failed on a write it never made.

It had simply never run in the same process as that spec. `scripts/spec_shard.sh` partitions by **file size**, so which specs share a runner is a function of the tree — the ~190 lines of specs added here re-packed the bins. Adding examples to any unrelated file would have done the same, which is why it is fixed at the source rather than re-run: `Settings.rewriter_rules` and `rewriter_next_rule_id` are now saved and restored around each example, the same pair the `settings_spec` helpers already capture.

Verified by running CI's exact partition locally — all four shards green:

```
shard 0/4  2971 examples, 0 failures
shard 1/4  2949 examples, 0 failures
shard 2/4  3032 examples, 0 failures
shard 3/4  2848 examples, 0 failures
```

PR #890, opened 2026-08-30
fix(cli): what the listings got wrong — a silent scan cap, two glued columns, a dropped port, and a dangling flow id

hahwul:
Five defects on the CLI's read surfaces. None of them crashes; every one of them
prints something an operator or a script then believes. Each commit is one fix
plus a regression spec that was confirmed red against the old code.

### 1. `sitemap` hit its scan cap silently — and then counted off the truncated set

`collect_sitemap` called `store.sitemap_entries(filter, limit, …)` and never
compared the returned size to `limit`.

```
$ gori run sitemap --project demo -n 5
acme-corporation-staging-….demo.test  (1 path)
…
# nothing on stderr, in text, json OR paths
```

The short host list is bad enough; the per-host `(N paths)` header is counted off
whatever survived the cut, so a capped run makes a *positive, wrong* claim about
the project. The cap counts 6-column transport keys — the same multiplication
`sitemap_node_exists?` already documents — so the 10k default is reachable on a
real engagement, not just with `-n 5`.

The cut is measured on the raw read (`entries.size >= limit`, the test
`Diff::Snapshot` uses) *before* any fold makes it invisible, and the notice goes
to STDERR so json/paths keep a pure STDOUT. `sitemap_truncation_notice` is a pure
`String?` helper for the same reason `tag_match_warning` is one.

### 2. The METHOD column ran into the one beside it

```
$ gori run history -q 'method:OPTIONS'
#87    OPTIONShttps api.demo.test*  -> 204
```

`ljust(7)` guarantees a separator only while the value is *shorter* than the
column. `OPTIONS` and `CONNECT` are exactly 7 and both are RFC 9110 registered and
routine (CORS preflight, tunnels), so the row is unsplittable on ordinary traffic;
`PROPFIND`, `CHECKOUT` and `VERSION-CONTROL` follow.

`pad_cell` keeps the value whole and grows the column by the one space it owes.
Not a clamp: the TUI's fixed 8-cell version is right for a screen (its own comment
says truncating `PROPFIND` would *be* the regression), but a CLI listing has no
geometry to defend and a script is reading that token. Both sites — the history
row and `authorize_target_text`.

### 3. The history row dropped the port, while `--format json` kept it

The location cell was `Url.location(row.host, row.target)` — host + target, which
carries no port — so every origin-form (HTTPS/CONNECT) capture printed its host
bare and two services on one host became the same cell:

```
https 127.0.0.1/service-A     # :19315
https 127.0.0.1/service-B     # :19316
```

json told them apart the whole time (it emits `port`, and its `url` goes through
`FlowRow#url`), so the two formats disagreed about the same flow. Visible on the
seeded demo as well: `legacy.demo.test:8080/` printed as `legacy.demo.test/`.

The cell now routes through `row.url` and chops the scheme the SCHEME column
beside it already printed. Deliberately not `:#{port}` pasted in here:
`FlowRow.url_of` is the single definition of a flow's absolute URL, non-default
port and IPv6 brackets included, and a second copy on that line is a copy that
drifts. An absolute-form target keeps its scheme, unchanged — those bytes are the
request line the client wrote.

### 4. `issues create --flow` accepted a flow id that does not exist

```
$ gori run issues create --db X -t t --flow 424242
Issue #21 created successfully.          # exit 0
```

`--flow 0` and `--flow -5` were accepted too. The listing, the markdown report and
the SARIF location then advertise `flow#424242` as evidence a reader is told to
open. The sibling refuses the identical id: `links add --ref flow --ref-id 999999`
→ `no flow with id 999999`, exit 1.

Checked against the store before the insert, with the sibling's wording and exit
code. `flow_row`, not `get_flow` — the row-only read answers "does this exist?"
without materializing both BLOBs. `<= 0` is a separate arm because the store has
no row there to disagree with, and `no flow with id -5` would send the reader
hunting for a capture instead of at their own argument. `parse_flow_id` itself is
left alone on purpose: fuzz/mine/sequence load the flow and already fail.

The comment at `issues.cr:124` claimed "Same rule --severity and --flow already
follow". Only `--cvss`/`--severity` did; that sentence is corrected.

### 5. `sitemap tag --list` wrote a raw tag into a TSV line

A tag is free text and can be set from the TUI or MCP as well as `--tag`, so a
newline in one split a single tag across two physical lines and a tab invented a
fourth column. The tree view already folds the same tag to `# multi·line`; the
TSV line goes through the same `term_safe`, and `--list` has no `--format json`
to fall back to.

---

**Verification.** Each spec was confirmed red against the pre-fix behavior first
(new helpers neutered to their old semantics, then restored — no canaries left in
the tree). `crystal tool format src spec`, `crystal build --no-codegen
src/main.cr`, and the full suite: **11906 examples, 0 failures, 0 errors**.

Dogfooded end-to-end against a seeded `demo` project in an isolated `GORI_HOME`,
including the `links add` wording comparison and the text-vs-json port disagreement.

### DIFF
--- a/spec/cli/run/authorize_spec.cr
+++ b/spec/cli/run/authorize_spec.cr
@@ -201,6 +201,16 @@ describe "gori run authorize — output" do
     lines[3].should contain("different")
   end
 
+  # Same `ljust(7)` shape as the history row: `OPTIONS` (a CORS preflight is exactly the kind
+  # of request an authz sweep is pointed at) ran flush into the URL, printing
+  # `#7     OPTIONShttps://acme.test/admin`.
+  it "text: keeps a space between a 7+-character METHOD and the URL" do
+    t = A::Target.new(7_i64, "OPTIONS", "https://acme.test/admin", [
+      trial("as-captured", A::Verdict::Baseline, 200, 10_i64, baseline: true),
+    ])
+    Gori::CLI::Output.authorize_target_text(t).lines[0].should contain("OPTIONS https://acme.test/admin")
+  end
+
   it "text: an enforced request is quiet, and an all-errored one says error" do
     enforced = A::Target.new(8_i64, "GET", "https://acme.test/me", [
       trial("as-captured", A::Verdict::Baseline, 200, 10_i64, baseline: true),
--- a/spec/cli/run/history_spec.cr
+++ b/spec/cli/run/history_spec.cr
@@ -67,6 +67,37 @@ describe "gori run history — CLI::Output rows" do
     rel.should contain("api.test/a")
   end
 
+  # The location cell was `Url.location(row.host, row.target)` — host + target, no PORT — so
+  # every origin-form (HTTPS/CONNECT) capture printed its host bare and two services on one
+  # host were the SAME cell. `--format json` told them apart the whole time (it emits `port`,
+  # and its `url` goes through `FlowRow#url`), which is exactly what makes the text list
+  # misleading rather than merely terse.
+  it "keeps the non-default port of an origin-form flow, so two services on one host differ" do
+    a = Gori::Store::FlowRow.new(
+      id: 1_i64, created_at: 0_i64, scheme: "https", method: "GET", host: "127.0.0.1",
+      port: 19315, target: "/service-A", status: 200, size: 0_i64, state: Gori::Store::FlowState::Complete)
+    b = Gori::Store::FlowRow.new(
+      id: 2_i64, created_at: 0_i64, scheme: "https", method: "GET", host: "127.0.0.1",
+      port: 19316, target: "/service-B", status: 200, size: 0_i64, state: Gori::Store::FlowState::Complete)
+    Gori::CLI::Output.flow_row_text(a).should contain("127.0.0.1:19315/service-A")
+    Gori::CLI::Output.flow_row_text(b).should contain("127.0.0.1:19316/service-B")
+    # The SCHEME column already says https; the cell must not repeat it.
+    Gori::CLI::Output.flow_row_text(a).should_not contain("https://")
+  end
+
+  it "leaves a default-port flow and an IPv6 literal spelled the way FlowRow#url spells them" do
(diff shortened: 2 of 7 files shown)


hahwul:
Closing: superseded by the consolidated main, not abandoned.

All five commits were cherry-picked onto the integration main and verified there
before this was closed:

| this branch | integration main |
|---|---|
| `e1df545e` say when `gori run sitemap` hit its scan cap | `c2ebc7e7` |
| `9b45b40d` fold a multi-line tag in `sitemap tag --list`'s TSV output | `0588523a` |
| `f8157ad3` keep a space between the METHOD column and the one after it | `4053345d` |
| `35deca59` refuse a dangling `--flow` on `issues create` | `1cfcc9d6` |
| `e7523da9` keep the port in the history text row's location cell | `6945141c` |

All seven touched files — `src/gori/cli/output.cr`, `src/gori/cli/run/issues.cr`,
`src/gori/cli/run/sitemap.cr` and the four specs — are byte-identical on both
sides (compared by blob hash), so nothing was dropped or reshaped in transit. The
twelve regression specs are all present and passing there, which was the failure
mode worth checking: a cherry-pick that takes the `src` change and leaves the spec
behind is the quiet one.

Verification on the integration main: 12017 examples, 0 failures. This branch
itself was last checked green against `a4f94295` by building the actual merge
result locally (11943 examples, 0 failures) — CI does not build the merge result,
only the branch.

The five fixes, for the record: a scan cap `gori run sitemap` hit silently while
still printing `(N paths)` counted off the truncated set; a `ljust(7)` METHOD
column that glued itself to the next one for `OPTIONS`/`CONNECT` and every longer
verb, in both the history row and `authorize_target_text`; a location cell that
dropped the port on origin-form captures while `--format json` kept it; an
`issues create --flow` that accepted a nonexistent, zero or negative flow id and
advertised it as evidence; and a raw tag written into `sitemap tag --list`'s TSV
line, where a newline split one tag across two rows.

PR #897, opened 2026-08-30
feat(fuzz): save and restore permanent runs

sebastianosrt:
## Summary

- persist complete Fuzzer runs and every request, wire request, response head, and response body byte in the project database
- add `Shift-S`, per-session run history, deletion, and automatic latest-run restore to the TUI while keeping ordinary runs in memory and uncapped
- add `gori run fuzz save/list/show/delete` without changing the existing ephemeral `gori run fuzz` behavior
- add MCP `save_results` plus saved-run list/get/delete tools with bounded, redacted reads and explicit sensitive-byte access

## Design

All three surfaces write through the shared `Fuzz::Persistence` seam in bounded batches. Persistence failures are reported as `save_failed` and never stop outbound traffic. Saved rows retain terminal status and frozen transport context, while active or partial saves are excluded from automatic TUI restoration and require explicit stale recovery before deletion.

The latest successful run for the initially selected Fuzzer session loads asynchronously when a project opens. Other sessions load on first selection, preserving startup responsiveness while retaining the existing explicit history picker for older runs.

## Verification

- `just build`
- `just test` — 12,090 examples, 0 failures
- `just benchmark-check` — 49 harnesses
- changed-file format check and `git diff --check`


hahwul:
Hi @sebastianosrt
Thanks, this is solid work — the persistence writer especially.

I pushed the fixes onto your branch instead of sending it back. The teardown waits (`quiesce_view`, `stop_all`, `Spool#close`) all ran on the render fiber with no deadline, so closing a tab mid-run froze the TUI until the target answered — those are bounded now. Also gave the spool a per-run disk ceiling, and made a rejected append report at the first rejection instead of at the end of the sweep. Rebased on main.

Merging. Thanks again!

hahwul (approved):
Approved.

PR #898, opened 2026-08-30
feat(history): add direct delete and clear shortcuts

sebastianosrt:
## What changed

- bind `d` in the History list to the existing confirmed delete action for the selected flow or marked set
- bind `Shift-X` to the existing confirmed clear-all action
- keep `D`/`X` as the danger-menu mnemonics and leave flow-detail deletion menu-only
- document both shortcuts in Help and cover their exact chords and scope in specs

## Why

History already supported deleting flows and clearing the project through the Space menu, but neither action had a direct list shortcut. These bindings reuse the existing confirmation, busy-database handling, and batch-target semantics rather than adding another deletion path.

Bare `d` is list-local delete while `Space → d` remains Discover; the explicit `D` danger mnemonic keeps those menu actions distinct. `Shift-X` matches the existing clear-history mnemonic and avoids the list's plain `x` selection binding.

## Validation

- `crystal build --no-codegen src/main.cr`
- `crystal tool format --check src spec`
- `crystal spec` — 12,066 examples, 0 failures
- `scripts/bench_check.sh` — 49 harnesses build
- `just build`
- focused History/Help/keymap/space-menu specs — 99 examples, 0 failures

Ameba still exits non-zero on the repository's documented pre-existing backlog (933 findings); it reports no new finding on the changed source lines.


hahwul (approved):

hahwul:
Hi @sebastianosrt 
Thank you for the contribution! This touches a core interaction, so I spent some time thinking it through.

To share the outcome first: I plan to expand history's Shift+X into a global feature. 
I'll handle that in a follow-up PR (e.g. Projects > Activities, Issues, and so on)

Merging this now. Thanks again!

PR #899, opened 2026-08-31
feat(tui): one chord clears a tab — ⇧X across all four clear-all verbs

hahwul:
## What

gori has four verbs that delete a whole tab's worth of project data. They answered three different things:

| verb | scope | menu before | chord before | after |
|---|---|---|---|---|
| `history.clear` | Body | `X` | `⇧X` (#898) | unchanged |
| `probe.clear` | Probe | `X` | — | `⇧X` |
| `authorize.clear` | Authorize | `X` | — | `⇧X` |
| `activity.clear` | ProjectActivity | `C` | `⇧C` | `⇧X`, menu `X` |

All four are `⇧X` now, with `X` as the space-menu letter, so the gesture is learned once.

## Why ⇧X and not ⇧C

What sits under the shift. Bare `x` is bound in **none** of the four clear-all scopes. Bare `c` is live in **all** of them — `capture.toggle` is `Scope::Global` (`verbs/core.cr`), and `probe.dismiss-selected` shadows it on the Probe list, which is the most-pressed key on that tab. A project wipe one shift above the triage key an operator holds down all day is the arrangement to avoid, not the one to standardise on.

`C` was not available as a menu letter either: History spends it on `history.columns` and Repeater/Fuzzer on Send to Comparer.

## Why this is not 0edc3c5b in reverse

0edc3c5b ("Take ↵ off the toggle and the wipe off a bare letter, in Probe") took Probe's clear off a **bare** `x`, and its objection had two halves:

1. one chip over, on the Rules sub-tab, `x` is a harmless enable toggle;
2. no Probe hint ever named the destructive reading — "of the two meanings the unadvertised one was the dangerous one".

A shifted chord answers (1) by construction, and the toggle has since moved to `Scope::ProbeRules` (`ProbeController#command_scope`), so the two can never resolve on one keystroke. **(2) is answered in this PR**: every one of the four chords is now named where it can be read before it is pressed — Help rows and body hints, below. The space-menu entries stay exactly where 0edc3c5b put them; the chord is added to that shape, not traded against it. Each comment that stated the old rule now states the new one and says this explicitly, so the file history reads as one argument rather than a reversal.

## The one real collision

`activity.clear-filters` held `mnemonic: 'x'`. With `activity.clear` on `X`, the Activity space menu would have read:

```
x  Clear filters      <- resets a filter
X  Clear activity     <- permanently deletes the agent audit trail
```

`clear-filters` moves to `'N'`, the letter `history.mark-clear` already spends on "Clear marks". `Registry#validate_menu_keys!` confirms it is free in that scope at boot.

## Advertisement

**Help sheet** (`tui/help_view.cr`): the Probe row and the Project row's new indented `activity` row name `⇧X`, and **Authorize gained a row at all** — `TAB_SECTION` already pointed its Shortcuts popup at OTHER TABS, which never mentioned the tab. (`:target` is still row-less; noted in the comment, left for another change.)

**Body hints**: all four name the key, resolved through `Hotkeys.binding_label` so a rebind reaches the line that advertises it.

Two places deliberately do **not** name it, rather than being silently skipped:

- **Authorize's running-run hint.** It is deliberately the two keys a run leaves meaningful, and "empty the queue" is not what to put in front of an operator watching one go out. `AuthorizeController#clear` already refuses mid-run with "^X to stop it first", so `⇧X` there is inert and points at the key that was wanted.
- **Probe's preview-focus and History's preview-focus hints.** Both are the abbreviated form of a list hint that already drops several keys; the full-width default branch on each tab names `⇧X`.

History's hint names `⇧X` but not `d` (#898's other key): one row versus the whole tab, and that line is already the longest on the tab.

## Specs

The specs assert the **dispatch**, not the declaration. `Verb::Keymap#lookup` is asked for the chord `Keybind.from_event` actually produces from a shift+X event, in each of the four scopes, plus the two `toggle-default` scopes that keep their own `⇧X`. A `Chord.new("X")` spelling satisfies every equality assertion and never fires — that is how a dead binding shipped once before — so `spec_helper` grows a `shift_chord` helper that round-trips through the real event path, and the existing hand-rolled one in `activity_spec` folds into it.

Two tautologies removed:

- `help_view_spec` asserted `Item#key`, which `build_rows` **replaces** via `Hotkeys.binding_label` whenever a `verb_id` is set (`help_view.cr:175`/`356`). It passed even when the chord changed. It now asserts the resolved label, with `""` as the fallback so a miss reads as empty rather than as the literal it was meant to check.
- A new `help_view_spec` case pairs each keyless OTHER TABS row against `binding_label` for its verb, so a hand-written chord name in a description cannot drift from the registry.

`activity_spec` also pins the property that made `⇧X` the right key: bare `x` is unbound in all four scopes, and bare `c` resolves to something that is not the wipe.

## Cross-scope reuse

`⇧X` is also `colormarker.toggle-default` and `rewriter.toggle-default` ("Enable/disable everywhere"). `Conflicts.overlap?` is `a == b` on the scope, so the reuse is deliberate and undetected by design. No code change; both sites gained a line saying so, and neither tab has a clear-all verb to be confused with.

## Also

`⇧C` is gone from the docs (`docs/content/guide/proxy.md` + `.ko.md`) and from three stale comments that named it. Two `s`/`l`/`c` comments in `verbs/activity.cr` and `project_controller.cr` said `c` narrows the feed — the actor filter has been `a` for a while; corrected while in the file.

## CHANGELOG

Activity is unreleased (latest tag is v0.4.0; #864 landed after it), so its entry's `⇧C` sentence is **edited in place** rather than given a "changed" entry for a key nobody has shipped. #898's entry keeps the `d` half and cross-references this PR for `⇧X`, so `Unreleased` does not describe the same key twice in two ways.

## Verification

```
crystal tool format --check src spec          # clean
crystal build --no-codegen src/main.cr        # clean
crystal spec                                  # 12147 examples, 0 failures
./scripts/bench_check.sh                      # all 49 harnesses build
crystal build --no-codegen scripts/seed_demo.cr
```

ameba: diffed against the HEAD copies of every changed file — no new findings (the src/ set is byte-identical; `spec/**` `Lint/NotNil` exclusions only apply in-tree, which accounts for the apparent delta).


hahwul:
Code review found 11 items. Acting on 8; declining 3 with reasons.

## Fixed (d7c30f0e)

**`authorize.clear` had no confirmation.** The real one. `AuthorizeController#clear` went straight to `@view.clear` — the only clear-all verb in the app that did — while `history_clear`, `probe_clear` and `activity_clear` have always raised a danger confirm. This PR put it on one keystroke and the CHANGELOG claimed all four ask first, which was false. It now confirms, names the request count (`"the queue"` understates the identity results that go with it), and toasts on an empty queue instead of opening a dialog.

**Probe's clear was a silent no-op on an empty list.** `return if @probe.empty?` with no toast. Menu-only, the empty case was self-evident; as an advertised chord it reads as a key that failed. Now says so, matching `activity_clear`.

**Hints named the chord in only one branch per tab.** `command_scope` answers `Scope::Probe` for the mode-off list and both preview focuses, and `history.clear` is `Scope::Body` regardless of the preview pane — so my "abbreviated form" rationale in the original PR body was wrong: those branches are just as one keystroke from an empty tab. Naming it in the default branch alone rebuilt the same unadvertised-key gap in three of Probe's four states and two of History's three. All of them name it now. The Probe comment no longer claims the placement is free; it says the eleven columns come out of `mode`/`filter`, which the space menu still names.

**Stale docs.** `guide/authorize.md` + `.ko.md` still read `Space` → `X` … `(menu only)`. Fixed in both locales.

**`guide/hotkeys` stated the opposite policy** — "**Ctrl** is for actions that must work while typing (INS) or that are **destructive**". Amended in both locales: Ctrl keeps INS and workbench run/stop; Shift carries the whole-tab wipes, with the `x`-not-`c` reason and the two obligations a destructive chord takes on (named where it can be read before it is pressed; asks first). Good catch — this is the one document a future contributor would have read instead of the verb-file comments.

**`help_view_spec` lost the literal-key coverage.** `shortcut_rows(nil)` still renders `item.key`, so the fallback path was reachable with zero coverage. One assertion back, phrased as "agrees with the resolved label" rather than a bare `eq("⇧X")` — which is what let the spec pass while the chord moved.

**`shift_chord` claimed a suite-wide convention it does not have.** Comment corrected: it now names what is left unswept (the ⇧J/⇧K/⇧F/⇧E/⇧T assertions in sitemap/diff/oast/sequencer/comparer/issues) and why those are lower risk — reorder and navigation verbs, where a dead binding is visible the first time anyone presses the key. Sweeping them is a separate change.

## Declined

**Help rows hard-code `⇧X` in the description.** True, but structural and pre-existing: OTHER TABS rows spend the key column on the tab name, so `verb_id` has nothing to resolve, and every chord in those rows is already a literal (`⇧S scope`, `^R run`, `⇧T all`). Making them rebind-aware needs description-level interpolation on `Item` — a real feature, not this PR. The limitation is stated in the spec comment, and the new spec pins the literal against the registry at the default binding, which catches drift-by-code-change.

**`binding_label`'s fallback fires on an explicit unbind.** Correct, and a genuine wart — but it is `binding_label`'s semantics, shared by every one of the ~15 existing call sites (`repeater`, `filter`, `follow`, `issue`, `intercept`, `mode`, …). My four new sites behave identically. Fixing it means changing the facade app-wide.

**Four `binding_label` calls per render tick.** Pre-existing pattern — `history_controller#body_hint` already made five. A resolved-label cache is worth doing and would help every hint in the app, but it is not this change.

## Verification after the fixes

```
crystal tool format --check src spec          # clean
crystal build --no-codegen src/main.cr        # clean
crystal spec                                  # 12149 examples, 0 failures
./scripts/bench_check.sh                      # all 49 harnesses build
crystal build --no-codegen scripts/seed_demo.cr
```

ameba: one new finding on a line I had just written (`Style/VerboseBlock` in the restored help spec) — fixed; the rest is byte-identical to the baseline.

PR #900, opened 2026-08-31
fix(cli): refuse the stray positional every `gori run` parser was dropping

hahwul:
## The gap

`OptionParser`'s default `unknown_args` handler is SILENT, so a parser that installs none discards every leftover word without a sound and the command reports success for work it did not do. Reproduced against the built binary before the fix:

    gori run project scope add -p zzz.test STRAY          → "Scope rule added successfully."
    gori run rewriter add -f a -v b STRAY                 → "Rule #1 added."
    gori run sitemap tag --host h --path / --tag t STRAY  → "Tagged h/: t"
    gori run capture --project p STRAY                    → proxy starts, STRAY never mentioned

## Re-measuring first: the gap is 17, not 37

The `p.unknown_args` grep reports 37 missing across 12 files. Verified per site, **20 of those are false positives** and the real gap is **17**. The grep misses three guard spellings:

| Missed by the grep | Sites | Why the grep didn't see it |
|---|---|---|
| `parser.unknown_args` | 12 | the receiver is named `parser`, not the block's `p` |
| `views_one_positional(parser, …)` | 5 | guard installed inside a shared helper, after the block |
| `parse_no_positionals(parser, …)` | 3 | same, `repeater h2` / `list` / `create` (already landed) |

So `views.cr` (grep: 6 missing) and `repeater.cr` (grep: 3 missing) were already **fully guarded**, and `colormarker.cr` / `rewriter.cr` / `project.cr` / `issues.cr` were each partly so. The per-method scan used instead of the line grep is in the spec as the permanent gate.

## Guarded here (17)

All 17 take **no positional at all** — kind 1 in the classification. None reads `args` again after parsing (checked mechanically), and every dispatcher strips its own verb before calling them, so a leftover word is genuinely stray.

| File | Method | Subcommand |
|---|---|---|
| `capture.cr` | `cmd_capture` | `gori run capture` |
| `colormarker.cr` | `cmd_colormarker_color_add` | `colormarker color add` |
| `colormarker.cr` | `cmd_colormarker_add` | `colormarker add` |
| `colormarker.cr` | `cmd_colormarker_preview` | `colormarker preview` |
| `decoder.cr` | `cmd_decoder_list` | `decoder list` |
| `intercept.cr` | `cmd_intercept_toggle` | `intercept enable` / `disable` |
| `issues.cr` | `cmd_issues_create` | `issues create` |
| `oast.cr` | `oast_listen` | `oast listen` |
| `probe.cr` | `cmd_probe_rule_add` | `probe rules add` |
| `project.cr` | `cmd_scope_add` | `project scope add` |
| `project.cr` | `cmd_sandbox_set` | `project sandbox on` / `off` |
| `rewriter.cr` | `cmd_rewriter_preset_list` | `rewriter preset list` |
| `rewriter.cr` | `cmd_extract_add` | `rewriter extract add` |
| `rewriter.cr` | `cmd_rewriter_bindings` | `rewriter bindings` |
| `rewriter.cr` | `cmd_rewriter_add` | `rewriter add` |
| `rewriter.cr` | `cmd_rewriter_preview` | `rewriter preview` |
| `sitemap.cr` | `cmd_sitemap_tag` | `sitemap tag` |

Each routes through the existing `Run.parse_no_positionals` rather than a hand-rolled copy: it installs `unknown_args { |before, after| … before + after }` in one place, so the `--` half cannot be dropped by a copy that keeps only `before` — the bug this surface has had fixed before. The shared parser helpers themselves are untouched.

Hints name the flag the value belonged to. On these commands the dominant real cause of a stray positional is an **unquoted multi-word flag value** whose tail words arrive as positionals, so the refusal prints every one of them:

    $ gori run colormarker add --when host:a AND status:500
    gori run colormarker add: unexpected arguments "AND status:500" — pass the condition as
    --when FILTER — quote it, an unquoted QL query splits into several arguments

## Deliberately left alone (20) — already guarded

Not touched; each already refuses or legitimately consumes its positionals.

- **Kind 3, takes positionals and already validates** — `views add/rm/rename/set/scope` (5, via `views_one_positional`, which refuses both missing and extra); `rewriter preset add`, `project scope update`, `probe rules enable/disable`, `probe rules delete`, `probe mode`, `session add`, `session edit`, `grpc reflect`, `grpc forget`, `oast resume` (via `one_positional` / `one_positional_list`).
- **Kind 1/2, already installs its own handler** as `parser.unknown_args` — `colormarker color update`, `colormarker color rm`, `colormarker list`, `colormarker rm`, `colormarker set_enabled`, `colormarker move`, `views list`, `rewriter extract_target`, `rewriter rm`, `rewriter set_enabled`, `project scope delete`, `issues update`.
- **Already on `parse_no_positionals`** from #897's neighbour work — `repeater h2`, `repeater list`, `repeater create`.

`src/gori/cli/run.cr` itself declares no `OptionParser`, so it is not in scope.

## Specs

`abort` calls `exit`, so a refusal is not catchable in-process — the repo's established shape (`list_leftovers_spec`, `option_parser_missing_option_spec`, `spec/cli_spec.cr`) is a source gate plus a decision-function spec, and that is what `spec/cli/run/unknown_args_sweep_spec.cr` adds:

1. **Repo-wide gate** — every `OptionParser` under `src/gori/cli/run/` must reach a guard. `repeater_positional_spec` scoped this to one file precisely because "~30 parsers elsewhere are still missing theirs, and a repo-wide gate here would be a red suite standing in for a sweep nobody has done". That sweep is this PR, so the gate moves here and widens; the repeater spec keeps its `no_positional_error` examples and its comment now points at the new home.
2. **Both-halves gate** — catches a handler that NAMES `before, after` and then reads only `before`. `spec/cli_spec.cr` already refuses the `|x, _|` spelling; that regex cannot see this one.
3. **Per-subcommand pairing** — each of the 17 must name its own subcommand in the guard it installs. The mistake a 17-site sweep is likeliest to make is not a missing guard but one carrying the neighbour's message, which no generic gate can see.
4. **`no_positional_error` examples**, including the `--` case driven through a real `OptionParser`: it asserts that `before` alone is empty (i.e. a `before`-only guard accepts `cmd -- stray` silently) while `before + after` refuses it.

Every gate has a negative control, and each was mutation-tested by actually breaking the source:

| Mutation | Caught by |
|---|---|
| guard deleted from `sitemap tag` | gate 1 + pairing (`sitemap.cr@23`, `sitemap.cr#cmd_sitemap_tag`) |
| `rewriter preview` guard renamed to `rewriter add` | pairing + its neighbour control |
| `notes.cr` handler reduced to `= before` | both-halves gate (`notes.cr:46`) |

## Verification

    crystal tool format --check src spec          OK
    crystal build --no-codegen src/main.cr        OK
    crystal spec                                  12153 examples, 0 failures
    ./scripts/bench_check.sh                      all 49 harnesses build
    crystal build --no-codegen scripts/seed_demo.cr   OK

ameba: 15 findings on the changed files, **0 new** — compared against the same files at `HEAD` by rule+method identity, since my added lines shift every line number.

End-to-end against the built binary, under an isolated `$GORI_HOME`:

- all 17 refuse a stray positional, exit 1, each naming its own subcommand;
- `decoder list -- stray`, `rewriter bindings -- stray`, `sitemap tag … -- stray`, `project scope add … -- stray` are refused too (the `--` half);
- **no regression**: all 17 valid invocations still succeed, including `capture` (binds, serves, stops on `--for`), `oast listen` (reaches its provider check), and quoted multi-word values (`--when 'host:a AND status:500'`, `--title 'a title with spaces'`, `--tag 'a memo with spaces'`).

## Note — unrelated defect, not fixed here

`src/gori/cli/run/sitemap.cr`: `sitemap_tag_row` exists with a doc comment explaining that printing a tag raw lets a newline split one tag across two lines and a tab invent a fourth column, desyncing host+path from the memo. **Production never calls it** — `sitemap tag --list` prints `puts "#{k[0]}#{k[1]}\t#{t}"` raw. `spec/cli/run/sitemap_spec.cr` is green because it drives the helper directly, so the spec stands over a path that does not use it. Left for its own change.


hahwul:
## Review follow-ups (fb06cd1d)

Addressed the findings that were defects in what this PR adds:

- **Both-halves gate passed vacuously on short parameter names.** It tested `String#includes?` on the block-parameter *names*, so `{ |b, a| positional = b }` was accepted — `positional` contains an `a`. Now matched as whole words, with a control pinning exactly that case. Confirmed by mutation: the old check let it through, the new one flags `notes.cr:46`.
- **Fixed 7-line look-ahead** on the same check gave a `do … end` handler longer than the window the wrong verdict in *either* direction. Now windowed to the block's own `end`, with controls for both a late `after` read and a late drop.
- **Non-recursive glob** (`*.cr`) where both sibling gates use `**/*.cr` — would exempt a parser added under any future subdirectory. Aligned.
- **Ordinal-pinned "three spellings" example** would have silently moved onto a different parser if one were inserted ahead of it. Anchored on the enclosing method instead.
- **`sitemap tag` hint named flags the `--list` form does not read**: `--list acme.test` was advised to pass `--path`/`--tag`. It now also names `--host`, the flag that would have worked.
- **CHANGELOG entry added**, marked as breaking and listing the affected subcommands.
- **Pairing-check comment** now states what it actually pins (the prefix, not the hint) and why hints are deliberately shared between `add`/`preview` siblings.
- Dropped the one `no_positional_error` example that restated `repeater_positional_spec`'s own.

Re-verified: format, build, **12,154 examples / 0 failures**, 49 bench harnesses, seed_demo, ameba 0 findings on the changed file. Re-ran the mutation tests against the strengthened gate, and re-checked the reworded hint against the built binary on both `sitemap tag` forms.

## Not fixed here — deliberately

- **`gori mcp` has no `unknown_args` handler** (`src/gori/cli/mcp.cr:23`) — confirmed: 1 parser, 0 guards, while `ca.cr` and `settings.cr` are guarded. It is a real instance of the same defect, but outside `src/gori/cli/run/`, this PR's scope. Widening the gate to `src/gori/cli/**` would turn the suite red on an unswept file — the "red suite standing in for a sweep nobody has done" the repeater spec warns about. Its own change.
- **`project sandbox enable STRAY` refuses as `sandbox on`** — accurate, but pre-existing: `action = enable ? "on" : "off"` already drives the `invalid_option` and `missing_option` messages three lines above. Matching them is what the sweep was asked to do; changing the verb the whole method reports is a separate change to all three handlers.
- **`sitemap_tag_row` still has no production caller** — the defect noted in the PR description. Unchanged here.

PR #907, opened 2026-09-01
fix(tui): wire hjkl into the Repeater's request/target/response read caret

ksg97031:
## Summary
- Arrow keys already moved the caret in every Repeater READ pane; `h`/`j`/`k`/`l` did nothing there, unlike History's equivalent read handlers, which already alias both.
- The request pane and target pane had no vim-key support at all; the response pane only had `j`/`k` wired from an earlier "j/k parity" pass, missing `h`/`l`.
- None of `h`/`j`/`k`/`l` carry a bare-key verb chord in the Repeater scope (their `mnemonic` only fires inside the space menu), so aliasing them onto arrow-key caret movement doesn't shadow any existing shortcut.

## Test plan
- [x] `crystal build src/main.cr --no-codegen -o /dev/null` compiles clean
- [x] `crystal spec spec/tui/repeater_view_spec.cr spec/tui/target_row_select_spec.cr spec/tui/repeater_ws_editor_parity_spec.cr spec/tui/repeater_decode_spec.cr spec/tui/read_chrome_badge_parity_spec.cr` — 233 examples, 0 failures
- [x] Remaining `spec/tui/repeater_*` suite — 116 examples, 0 failures

### DIFF
--- a/src/gori/tui/controllers/repeater_controller.cr
+++ b/src/gori/tui/controllers/repeater_controller.cr
@@ -2535,18 +2535,18 @@ module Gori::Tui
       c = ev.char || key.to_char
       selecting = ev.shift?
       case
-      when key.enter?     then view.enter_request_insert!
-      when c == 'i'       then view.enter_request_insert!
-      when word_step?(ev)                  then view.request_read_move(0, key.left? ? -1 : 1, selecting: selecting)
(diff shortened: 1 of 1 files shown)


hahwul (approved):
LGTM

hahwul:
@ksg97031 
Thanks for contributing and Great work! Merged

PR #909, opened 2026-09-01
fix(mcp): refuse a send_request that names two requests (#906)

### LINKED ISSUE #906: MCP send_request replays the original Repeater request before reporting ignored overrides
## Summary

When `send_request` receives a `repeater_id` together with request overrides such as `url`, `method`, or `body`, the overrides are ignored and the stored Repeater request is sent unchanged.

The response reports `ignored_fields` and a precedence warning, but only after the network request has already executed. For state-changing requests, this can repeat an unintended operation before the caller learns that its overrides were discarded.

## Generic reproduction

1. Save any state-changing request as Repeater `R`.
2. Call:

```text
send_request({
  repeater_id: R,
  url: "https://example.invalid/different-path",
  method: "POST",
  body: "replacement body"
})
```

3. Observe that Gori sends the original request from Repeater `R`.
4. The result then reports that `url`, `method`, and `body` were ignored.

No credential values are needed to reproduce this behavior.

## Actual impact observed

During an audit workflow, the caller attempted to derive a harmless validation request from a saved Repeater while replacing its destination and body. Gori instead replayed the original state-changing request, creating an additional server-side resource.

This is more than a usability issue: the warning arrives too late to prevent side effects.

## Expected behavior

One of the following would be safe and predictable:

- Reject conflicting arguments before opening a socket.
- Apply the supplied overrides to the stored Repeater request.
- Require an explicit flag such as `override_repeater: true` before applying mixed input.

The current precedence rule should not silently reach the wire and explain itself afterward.

## Suggested fix

- Validate argument conflicts before building or sending the request.
- Return a structured `INVALID_ARGUMENT` error naming the conflicting fields.
- Document the precedence behavior in the MCP schema.
- Add tests asserting that rejected combinations perform zero network sends.

## Related MCP consistency gaps

These were found in the same workflow and are not covered by #904:

- `create_repeater` cannot set `tags`, while `update_repeater` can.
- `update_repeater` cannot patch only the path, headers, or body; callers must replace the entire raw request.
- Issue lifecycle fields cannot distinguish source-level latent risk from runtime states such as `unverified`, `not-reproducible`, or `inactive`.
- `list_history` lacks server-side aggregation such as unique method/path routes and status counts, so large coverage queries return oversized JSON arrays that may be truncated.

The first item in this report is the priority because it can cause unintended side effects.

hahwul:
Closes #906.

## The defect

`send_request` accepted a stored source (`flow_id` / `repeater_id`) together with the arguments that define a request, ignored the second set, sent the stored bytes, and then explained itself:

```json
{"status": 201, "ignored_fields": ["url","method","body"],
 "precedence_warning": "flow_id takes precedence; url, method, body were ignored"}
```

The precedence itself was defensible. Its *position* was not. The report is assembled in `send_result_json`, which runs after `plan.send_wire` — so for the case the issue reports (an agent deriving a harmless validation request from a saved state-changing one by replacing its destination and body) the original POST executes a second time for real, and the warning arrives after the side effect. A warning cannot un-create the resource it describes.

## What changed

**One gate, first statement in `send_request`.** `send_source_conflict` refuses a call whose arguments name more than one request, ahead of every other read — ahead of the History row `record_history` writes, and ahead of the dial:

```json
{"error_code": "INVALID_ARGUMENT", "field": "url",
 "details": {"conflicting_fields": ["repeater_id","url","method","body"]}}
```

The text says `NOTHING was sent` and names both exits: drop the overrides to send the stored request, or drop the id and describe the request yourself (with `get_flow` / `get_repeater_context` named as the way to read the stored bytes to start from).

The two pre-existing pre-send refusals in this path — `flow_id` + `repeater_id`, and `h2_fields` + a source — moved into the same gate. They were bare `Gori::Error` raises, so they reached the caller as `INVALID_ARGUMENT` with no `field` and no `details`; they are now the same coded shape as the new one, and `send_plan_options` / `field_native_plan_options` no longer re-check what the gate has already settled.

**Rejection, not application.** #906 lists three acceptable shapes; this is the one its own *Suggested fix* names, and it is the house rule for this exact pair already (`fuzz_start`'s "pass ONE template source", `gori run fuzz`'s "`<flow-id>` and `--flow/--repeater/--request` cannot be combined"). It is also the direction that stays open: `gori run repeater <flow-id> -H/-b/--target` *does* apply overrides onto a capture, so a later release may choose to bring that to MCP, and a caller written against the refusal keeps working. The reverse is not true — teaching agents that the pair means "my overrides are ignored" and later honouring them silently re-aims live traffic.

Per-send **modifiers** are unaffected and explicitly excluded from the conflict set: `http2`, `sni`, `tls_preset`, `timeout_ms`, `insecure`, `reframe_grpc`, `keep_request_line`, `verbatim`. Modifying how the stored request is sent is what they are for. There is a spec asserting this, so the refusal cannot be implemented by refusing everything.

**`ignored_fields` / `precedence_warning` are gone** from the response — the combination that produced them can no longer reach the wire.

## A second silent-ignore in the same seam

Auditing what else `send_plan_options` accepted and dropped turned up one: **`verbatim` never reached a `repeater_id` replay.** It was read only inside the `raw` branch, so `send_request{repeater_id, verbatim:true}` expanded `$NAME` and resynced Content-Length regardless — while `gori run repeater send --verbatim` (`CLI::Run::Repeater.session_plan_options`) and `send_websocket{repeater_id, verbatim}`, one screen up in the same file, have always honoured it on a stored session.

Same defect class as the issue — an argument accepted by name and dropped on the way to the wire — but the cost is the request's payload rather than its destination: a session whose stored body **is** `{"$where":"1==1"}` was sent with a project env var substituted into it and reported as a clean send of bytes the operator never wrote, or refused outright for an unbound variable nobody typed.

The reading is hoisted next to `keep_request_line` and `reframe_grpc` (which carry that "read up front, apply to all three branches" comment for the same reason) and applied to all three:

- `repeater_id`: `expand_request`, `preserve_field_case`, `auto_content_length` — `session_plan_options` exactly.
- `flow_id`: a capture is already byte-exact without the flag, so field case on h2 is the only thing left for it to promise; wiring it means the argument is never accepted and dropped on either stored-source path.
- `raw`/`url`: unchanged behaviour, now reading the hoisted value instead of its own.

## Verification

`crystal spec` (12,267 examples, 0 failures), `crystal tool format --check src spec bench scripts`, `crystal build --no-codegen src/main.cr`, `scripts/seed_demo.cr`, `scripts/bench_check.sh` (49 harnesses), `scripts/nix_shards_check.cr`. ameba on the two touched files: 41 findings before, 41 after — no new ones.

Eight specs, and three of them were mutation-checked rather than trusted:

- reverting the gate call fails all four refusal specs;
- reverting `preserve_field_case` on the flow path fails the h2 field-case spec;
- reverting the three `verbatim` flags on the repeater path fails the `$where` spec.

"Zero network sends" is asserted structurally, not by absence of an error: the refusal specs assert `store.recent_flows` is unchanged (the History row is written *before* the dial, so a row would prove the send path was entered), and the `repeater_id` one then sends the same repeater cleanly and gets 200 back from the **one-shot** origin — which is only still there to answer because the refused call never consumed it.

## On the other items in #906

- `create_repeater` **can** already set `tags` — #908 (2db74dcd) closed that one; no change needed.
- Partial `update_repeater` patching (path/headers/body), issue lifecycle states that separate source-level latent risk from runtime states, and server-side aggregation for `list_history` are each a feature rather than a defect. They are not touched here and are worth their own issues.


hahwul:
Review pass applied — 10 of 14 findings were real and are fixed in 4c1fb81f; the rest are recorded below with why.

**Fixed in the refusal itself**

- `field` now names the **source**, not an override. This was the sharpest one: `field` is what a client rendering a single argument tells the model to remove, and removing an override is the resolution that *re-sends* the stored request — the exact side effect this PR exists to prevent.
- A doubled conflict (`{flow_id, repeater_id, url, body}`) reports both halves in one `conflicting_fields` instead of refusing twice for a different reason each time.
- An empty `url` / `headers` / `body` no longer counts as an override. A client that fills every declared property named no second request, and `RequestBuilder.verbatim?` already reads an empty `raw_base64` that way — a gate on `present?` and a builder on `.presence` would have disagreed about one call. `describes?` is that reading, named once beside `present?`.
- The `h2_fields` conflict gets its own remedy (`url` + `h2_fields`). The generic one prescribed the h1 text carrier, which is precisely what an `h2_fields` caller is bypassing.
- Both remedies now name `include_sensitive:true` and `request_read_more` — `get_flow` / `get_repeater_context` redact `Authorization`/`Cookie` and cap at 16 KiB, so the prescribed way out was quietly lossy.
- `save_as_repeater` on a verbatim send stores `auto_cl: false`, matching the send. This one was newly introduced here: the row would otherwise re-frame a deliberate `Content-Length: 3` on its first replay, so the saved tab could not reproduce the response stored beside it.
- `field_native_plan_options` validates `verbatim`, so `{h2_fields, verbatim:"yes"}` is no longer swallowed on the one branch that never read it.
- `keep_request_line` dropped from the "combines with a source" enumeration — it is `flow_id`-only, and my new schema sentence had claimed otherwise.

**Fixed beyond `send_request`**

`mine_start` and `sequence_start` carried the precedence `fuzz_template_source` was already fixed for ("This used to return on the FIRST of template → flow_id → repeater_id") — returning on the first seed and never saying the other was dropped, on two tools that send for real, against CLI siblings that abort on the identical pair. Refused before a job exists.

**Specs**

The modifier example asserted nothing about any modifier; the session is now stored as h2 against an h1-only origin, so a 200 proves `http2:false` was threaded rather than merely allowed past the gate. Added the h1 half of the verbatim promise (bare-LF terminator preserved with the flag, promoted without it), plus examples for the empty-argument reading, `field` targeting, the `h2_fields` remedy, the saved-row `auto_cl`, and the mine/sequence guard. The mine/sequence guard was mutation-checked like the others.

**Filed, not fixed: #910**

`verbatim` does not stop a **declared binding** (extract rule / session slot) from being substituted at the send seam — `expand_request: false` only stops the builder pass; `Sender#wire` runs `Env.expand_bindings` unless `evidence: true`. Real, and `spec/send_seam_provenance_spec.cr:284-292` demonstrates the mechanism. But it predates this PR and affects `send_request{raw, verbatim}` and `gori run repeater send --verbatim` identically, project env vars — the `$where`/`$filter` case this PR's CHANGELOG line describes — go through the builder pass and *are* stopped, and resolving draft-bindings vs. verbatim is a DESIGN.md §7 decision that has to land on all three surfaces at once. Half-fixing one surface is how this seam drifted the first time.

**Not changed**

A refused call still writes an `agent_action` row to the event feed. Every argument refusal in this tool already does, and `log_agent_action`'s own comment wants a refused or scope-blocked send visible to the operator; the `TOOL_DISABLED` exclusion is about a tool that was never enabled, not about a call that was answered. This PR adds one refusal to an existing class rather than creating a new one.

Gates re-run green after the fixes: 12,274 examples, format, build, `seed_demo.cr`, `bench_check.sh`, `nix_shards_check.cr`. ameba on the six touched files: 51 before, 51 after.

PR #922, opened 2026-09-02
feat(tui): Settings › Language — English + 한국어, per area, applied live

hahwul:
Stacked on #921 (hahwul/tui-i18n-prework). The core of the TUI i18n series: the catalog module, the settings section, live apply, a LANGUAGE wizard step, and Miss Ring's own lines. English stays the default; the user picks in Settings › Language or on the wizard's first step.

**Shape**

- `Gori::I18n` (src/gori/i18n.cr): msgid-style — the English string in code *is* the key (`I18n.ui("History")`, `I18n.sys("sent → %{host}", host: h)`, `I18n.ui_n(n, "%{n} flow", "%{n} flows", n: n)`). English is the identity locale (no en catalog), a missing entry falls back to the msgid, a placeholder the caller did not pass falls back to the English template, nothing on the render path raises. Translation happens where text is *drawn*, so registries and tables stay English and a switch is a repaint; `I18n.revision` is for the few caches that bake text.
- **Four domains** — `ui`, `help`, `system`, `companion` — each resolve their own language: a default (`en`, `ko`, or `auto` from `GORI_LANG` → `LC_ALL` → `LC_MESSAGES` → `LANG`) plus per-area overrides that `inherit` it. Settings section `language` mirrors `companion.cr` line for line.
- **Only the TUI installs it**: `Tui.apply_language` at boot and on save. `gori run` / `gori mcp` never call it, so their output stays English and a core module that phrases a sentence through `I18n` (peer notices, in #D) is still English headless.
- `src/gori/i18n/locales/ko.json` is embedded at compile time (the `Fuzz::Presets` pattern). A seam spec (`spec/i18n_catalog_spec.cr`) holds it to the source in both directions: every ko entry must be a msgid something wraps (a reworded literal cannot orphan a translation) and every wrapped msgid must have an entry (nothing ships untranslated). Placeholders and `^X` key tokens must survive translation.
- Settings UI draws endonyms (`English`, `한국어`) while saving codes; a save applies on the next frame (Help rows rebuilt, like a keymap change); factory reset and the project picker apply it too.
- The setup wizard now opens on a LANGUAGE step whose live preview is the wizard itself; esc reverts. The wizard's own copy is the first surface drawn through `I18n`.
- Miss Ring's eight authored lines speak the companion language; what she relays keeps the system one.
- Docs: config reference + settings guide (en + ko), `docs/i18n/GLOSSARY.md` (register, what stays English, term table pulled from the existing `.ko.md` corpus), a DESIGN.md §7 entry.

**Verified**: `just test` (12,393 examples), `just benchmark-check`, format check; a headless pty run of `gori wizard` showed the Korean preview on ↓, the Korean NETWORK step, English again on ↑, and no `language` key persisted after esc.


hahwul:
Shelved after review: the per-string upkeep (every new TUI string needs a Korean line to pass the catalog gate) outweighs the payoff at this project's change rate. #921 keeps the layout/toast/settings fixes, which stand on their own. The commits stay reachable from this PR if the idea is picked up again.

PR #923, opened 2026-09-02
feat(tui): the interface in the chosen language — chrome, Help, settings, overlays, hints, confirms

hahwul:
Stacked on #922 (hahwul/tui-i18n-core). The interface sweep: every table-driven and chrome string is drawn in the interface language, and the Help/explanatory ones in the help language.

- **Table homes, translated at the draw site, English kept in the table and in every search haystack** so `/` and the palette answer to either spelling: verb titles (palette, space menu, hotkeys editor) and descriptions (hotkeys footer), the Help tab's sections and items and its Query page (QL + InterceptFilter tables, read at build time and rebuilt on `I18n.revision`), settings labels/hints/sections/groups (with choice labels), the space-menu bands, the focus badge and the shell's key hints, the copy/send menus, the activity chip, the empty-state titles.
- **Overlays**: every `title`/`hint`, the picker titles handed in at their open sites, import/export blurbs.
- **Per-tab body hints** in all 25 controllers — chord fragments ride as `%{}` placeholders so a rebind still reaches the translated line.
- **Confirm dialogs**: heading and buttons as interface text, the body as a system message; hand-wrapped bodies collapse to one msgid each, since `ConfirmDialog` wraps by display width. The confirm-wording guard now reads the wrapped literals.
- Decisions as approved: tab and tool names stay English (as in the docs); everything around them translates.

The seam spec enumerates each table so the Korean catalog is complete by construction (~1,900 entries after this PR).

**Verified**: `just test` (12,393 examples), `just benchmark-check`, format check.

PR #924, opened 2026-09-02
feat(tui): system messages in the chosen language — toasts, notifications, prompts

hahwul:
Stacked on #923 (hahwul/tui-i18n-ui). The system-message sweep: what gori says back — toasts, flashes, notifications — in the system language.

- **~590 producer sites** (`status(…)`, `@toast = …`, `toast(…)`, `set_flash(…)`, `notifications.push(…)`) wrapped in `I18n.sys`, interpolations converted to `%{}` placeholders, `#{n == 1 ? "" : "s"}` plurals to `I18n.sys_n`; the 22 sites with nested ternaries by hand, with the inner fragments translated too.
- **Core-authored sentences** through `I18n` where they are phrased: `PeerNotices` (already the one builder shared with the headless `Log`, which stays English because only the TUI applies a language), `Session::ReconcileResult#summary`, the five `PlanError` ladders in the TUI adapters, the quit/leave prompts and the arm hint.
- **A refusal is a flag, not an English prefix.** `SettingsView#render_footer` coloured its status by `starts_with?("invalid")`; it now carries `@status_warn`, like `PreferencesView` already did.
- The space menu's batch titles (`Delete 3 flows`, `Forward 2 held messages`, …) are `I18n.ui_n` calls instead of `"%s"` tables + a plural helper.
- Left English on purpose: `Outbound`/engine refusal constants (compared by `==`, written to the wire), `ex.message` passthroughs, the durable activity feed, CLI/MCP output, and the proxy's trust-store diagnostic (also printed to STDERR headless).
- Known limit (by design, P0): the notification ring stores rendered text, so notes pushed before a live language switch keep their language.

**Verified**: `just test`, `just benchmark-check`, format check.


hahwul:
Shelved after review: the per-string upkeep (every new TUI string needs a Korean line to pass the catalog gate) outweighs the payoff at this project's change rate. #921 keeps the layout/toast/settings fixes, which stand on their own. The commits stay reachable from this PR if the idea is picked up again.

PR #973, opened 2026-09-05
fix(history): keep large-project filtering responsive (#967)

### LINKED ISSUE #967: History filter freezes the tui
I think the history filtering can be optimized, It freezes the tui in medium/large projects.

sebastianosrt:
Fixes #967.

## Why it freezes

History ran SQLite searches on the UI fiber. The 110 ms typing debounce reduces how often a query starts, but it cannot bound a single scan: `LIMIT 1000` limits returned rows, not the rows SQLite must examine. Host-prefix completion has the same problem. Moving either call into a fiber alone would still block Crystal's single-threaded scheduler.

## Changes

- Add opt-in Store query controls: an exclusively checked-out read connection yields through SQLite's progress handler (1,000 VM instructions between checks, with a 4 ms cooperative slice). Cancellation is connection-scoped; the handler is removed before the connection returns to the pool. The writer never inherits it, and Store shutdown cancels and drains controlled reads before closing connections.
- Give History one search worker and one host-completion worker, each with at most one replaceable pending request. Query/scope/view changes invalidate obsolete generations immediately. Capture updates coalesce into one follow-up refresh without repeatedly restarting unfinished work.
- Keep previous rows navigable with a searching indicator, anchor replacement results to the current selection, and distinguish query errors from empty results. Host suggestions remain cached and asynchronous, including the opt-in completion dropdown.
- Cover review findings with regressions: captures arriving during an unfiltered reload; sorting while old rows remain visible; deferred clicks versus later navigation; committed deletes/clear versus stale snapshots; completion refresh starvation; and shutdown when several reads share a cancellation token.

Intentional UI change: searches publish asynchronously, showing the previous result set while pending. Failed searches retain those rows with an error indicator. QL predicates, full-project search coverage, result limits, raw-byte handling, and CLI/MCP's default synchronous behavior are unchanged. No schema/index change or multithreading flag is needed.

## Measurements

`bench/history_filter_bench.cr` compares synchronous and cooperative queries over synthetic on-disk projects, checking result IDs while recording query duration, the maximum gap in a 1 ms heartbeat, allocations, cancellation latency, and capture-writer throughput. It covers metadata, short body terms, FTS, headers, regex, and combined scope/provenance filters.

On a 500,000-flow fixture with approximately 1 KB bodies, an absent body regex (`body~absentneedle`, three repetitions) gave:

| Metric | Synchronous | Cooperative |
| --- | --- | --- |
| Total query time | 2.72–3.03 s | 2.79–3.12 s |
| Maximum scheduler gap | 2.72–3.03 s | 9.36–11.00 ms |
| Allocations per query | ~688 MB | ~688 MB |

The final rebased commit was also measured with 100,000 flows on Linux x86-64, Crystal 1.21.0 and SQLite 3.53.4. For the same regex, synchronous scheduler gaps were 509–539 ms versus 9.18–9.68 ms cooperatively; total query times were 509–539 ms versus 498–522 ms, with ~137.6 MB allocated in either mode. Cancellation completed in 0.04 ms. A 100-flow capture burst during a no-match short-body search finished at ~495 flows/s synchronously versus ~15,314 flows/s cooperatively. These are synthetic fixture measurements, not a general capture-throughput guarantee.

This improves responsiveness rather than making every full scan faster. A single regex callback or filesystem operation remains non-preemptible, so the slice is not a hard latency guarantee.

Reproduce without `-Dpreview_mt`:

```sh
crystal build bench/history_filter_bench.cr -o bin/history_filter_bench --release
BENCH_SIZES=10000,100000,500000 BENCH_BODY=1024 BENCH_REPS=3 bin/history_filter_bench
```

## Validation

Rebased onto current upstream `main` (`0e4c74c7`) before the final build and suite.

- `shards build`: passed.
- `just test-changed upstream/main`: selected and ran the entire suite because `store.cr` is shared — **12,922 examples, 0 failures/errors**.
- `crystal tool format --check src spec bench scripts`: passed.
- Dependency consistency check: passed.
- `just benchmark-check`: all 52 harnesses passed.
- `just lint-gate upstream/main`: no changed file gained findings (15 Crystal files).
- `just check`: format and dependency checks passed; the full-tree ameba run still reports the existing 940 findings. The PR diff gate above passes.

The full suite also prints existing broken-pipe diagnostics from fake origins in unrelated fuzz connection/race specs; they do not fail the suite.


hahwul:
@sebastianosrt 
There was some flickering when a capture happened with a filter applied, so I made a small fix.
Thanks for contributing and Great work :)

FYI, v0.5.0 is planned to be released over the weekend

hahwul (approved):

PR #1004, opened 2026-09-09
feat(cli): read a repeater request from stdin, and name a second source (#1001)

### LINKED ISSUE #1001: feat(cli): accept raw Repeater requests from stdin
## Problem

Creating a Repeater session from a generated or large raw HTTP request currently requires either interpolating the entire request into --request-raw or writing a temporary file for --request-file.

For large Base64/binary-derived headers this is cumbersome and can expose request material through shell argument handling or process listings. It can also run into command-line length limits.

## Proposal

Add a mutually exclusive --request-stdin option:

    generator | gori run repeater create --target https://example.invalid --request-stdin

The input should preserve the raw HTTP request bytes/line endings consistently with --request-file.

No target, project, credential, or environment-specific information is included in this report.

hahwul:
Closes #1001.

`gori run repeater create --request-stdin` takes the raw request from a pipe:

```sh
generate-request | gori run repeater create --target https://example.invalid --request-stdin
```

Byte-for-byte what `--request-file` reads from a file, which is the issue's stated
requirement. Both `IO#gets_to_end` and `File.read` are an `IO.copy` into a
`String::Builder`, so CRLF framing stays CRLF and a body that is not valid UTF-8 arrives as
its own octets — verified end to end, not just reasoned about: a 72-byte request carrying
CRLFs, a NUL and `ff fe 01 02` stores the same sha256 through both doors.

## Why a flag, and not `--request-file=-`

`read_input_file` already understands `-` behind an opt-in `stdin:` keyword, and the comment
at `src/gori/cli/run.cr:447` says that opt-in is deliberate: the flags that reject `-` as an
unreadable path should go on rejecting it rather than quietly blocking on a terminal read. So
`--request-file` is left alone and stdin gets its own named door, rather than two spellings
for one thing.

**No `STDIN.tty?` guard**, unlike `fuzz_source` / `mine_source` / `sequence_source`. Their
stdin road is *implicit* — the fallback taken when no source flag was passed — so without the
guard a bare `gori run mine` would hang on a terminal. This flag was named by the operator, so
blocking until EOF is the answer to what they asked for, and an interactive paste ending in
`^D` is a real use. It does get a **notice** on a tty, though: without one a forgotten pipe is
indistinguishable from a hung command, and every other interactive read in gori announces
itself first (`gori ca`).

## Two intentional behavior changes

**A second request source is refused.** "Mutually exclusive" cannot be implemented as a
stdin-only check without leaving an arbitrary hole, so the gate covers all three sources — and
that names a drop that was already there. The branch that reads the request is an `if`/`elsif`
chain, so:

```
$ gori run repeater create -t http://h --request-file req.txt --request-raw 'GET / HTTP/1.1'
Repeater session #8 created successfully.     ← the FILE was stored, the string never mentioned
```

Two sources cannot both be the request, and picking one by parser order is a guess made
silently. `--flow` is deliberately *not* one of the sources: it doubles as provenance for a
hand-authored request (`flow_id` on the row), so `--flow 42 --request-stdin` is a legitimate
pair.

**An empty request is refused, from every source.** A session whose request is empty cannot be
sent (`PlanError::NoRequest`), and MCP's `create_repeater` already refuses one — so this closes
a surface-parity gap rather than inventing a rule. `--request-raw ''`, an empty
`--request-file` and a generator that died each used to buy a dead row reported as a clean
"session #N created successfully". The message names the culprit:

```
gori run repeater create: the request must not be empty (--request-stdin gave no bytes)
```

Same shape and reason as `intercept edit`'s existing empty-replacement refusal.

## The trap

`repeater.cr` seeded the request from the flow with `if request_file.nil? && request_raw.nil?`
— a hand-extended chain, three lines under a comment warning that "an explicit request must
NOT be silently overwritten by the flow's bytes." A third source added to that chain by hand
is a `--flow N --request-stdin` whose piped request is overwritten. The source list is now one
function, `request_sources`, built once and shared by the gate and the seeding test. Verified
against the stored rows:

| invocation | stored request | `flow_id` |
| --- | --- | --- |
| `--flow 1 --request-stdin` | `PATCH /piped-wins …` | 1 |
| `--flow 1 --request-raw 'DELETE /raw-wins …'` | `DELETE /raw-wins …` | 1 |
| `--flow 1` | `GET /orig …` (the capture) | 1 |

## Ordering: every argv-only refusal sits above the blocking read

`--request-stdin` blocks until EOF, so a refusal knowable from the arguments alone has to be
reported before the read or the command drains the pipe first — and hangs outright on a
terminal — before saying something it knew all along. Two were on the wrong side: the source
conflict by construction, and `--target is required` for the whole life of the command
(`gori run repeater create --request-stdin` with no target hung on a fifo until killed;
now it refuses immediately). The read itself stays **above `open_store`**, so a pipe that never
ends is not holding the project's shared `<db>.open.lock` while it waits.

## An unreadable stdin no longer backtraces

`gori run repeater create -t http://h --request-stdin 0<&-` (fd 0 closed — a cron/systemd
unit, a `Process.run` with no stdin pipe) raised `IO::Error`, which `Run.dispatch` re-raises
and `CLI.run` does not catch (`Gori::Error` only), so the operator got a Crystal backtrace.
That is precisely the class `read_input_file`'s `File::Error` rescue exists to prevent for the
file twin; the stdin door now has the same guard:

```
gori run repeater create: cannot read the request from stdin: Closed stream
```

## Verification

`spec/cli/run/repeater_request_source_spec.cr`, 20 examples. The three verdicts
(`request_source_error`, `request_content_error`) and the branch selection (`request_content`)
are public functions rather than an inline chain, so the spec needs no test-only reopen of
`Gori::CLI::Run` — same precedent as `two_targets_error` — and it pins the **wiring**, not only
the predicates. Six mutations, each caught:

| mutation | caught by |
| --- | --- |
| unwire the empty-request refusal | the `PlanError` reason spec |
| unwire the source gate | the source-ordering example |
| empty check goes back to stdin-only | `request_content_error` examples |
| `--flow` seeding re-derives its own `.nil?` chain | the derivation example |
| delete the `--request-stdin` arm | 2 byte-identity examples |
| `.scrub` the piped bytes | the invalid-UTF-8 example |

The three refusals whose only home is a command ending in `abort` are asserted over the
source, the house pattern for exactly that (`interrupt_exit_status_spec.cr`: "both halves end
in `exit` and neither can be exercised in-process"). The `IO::Error` rescue is verified e2e
only — `abort` exits the process, so a spec cannot observe it.

`just test` 13260 examples / 0 failures · `crystal tool format --check src spec bench scripts`
clean · `just lint-gate` no changed file gained findings · `just benchmark-check` 56 harnesses
· `nix_shards_check` ✓.

One note on ameba: `cmd_repeater_create`'s pre-existing `Metrics/CyclomaticComplexity` finding
moves (threshold 12, already over on `main`). The finding count is unchanged, which is what the
diff gate measures; extracting the four helpers is what kept it from climbing, and
restructuring a 200-line command did not belong here.

## Surfaces touched

Flag + help description, the `parse_no_positionals` hint, the `gori run repeater` dispatcher
banner, both `docs/content/reference/cli.md` and `cli.ko.md` (tables, example, and a note on
the two behavior changes), one CHANGELOG line. No MCP or TUI counterpart is owed — MCP already
takes `raw` in the args hash, which is the same no-`ARG_MAX`/no-process-listing road, and
already refuses an empty request.


hahwul:
### Review round: what was declined, and two pre-existing defects left out of scope

**Declined — the sibling silent-precedence sites.** A review flagged `fuzz`, `mine` and
`sequence` as sharing the "picks one source by parser order" defect. They do not:
`fuzz` already refuses every pair of its three sources (`sources_given > 1` →
"pick ONE template source", `cli/run/fuzz.cr:212`), and `mine`/`sequence` each refuse their
one pair. `repeater create` was the only site, which is also why the sentence hardcodes the
command name rather than parameterizing it the way `two_targets_error` does.

**Declined — treating an empty flag value as unset.** `--request-raw "$UNSET" --request-stdin`
is reported as a conflict rather than silently taking the pipe. `two_targets_error` uses
`.presence` for its own question so that no working argv starts failing, but `--request-raw ''`
alongside a brand-new flag was never working argv, and ignoring an argument the operator did
type is the same class of silent drop this PR is closing. An empty `--request-raw ''` on its
own now gets the empty-request refusal, which says more than "required" would.

**Declined — `--request-file -` instead of the flag.** #1001 asked for `--request-stdin` by
name. The reviewer's strongest argument for `-` was that it would inherit `read_input_file`'s
rescue for free; the stdin door now has that rescue, so the argument is spent.

---

Two defects found during the review that are **pre-existing and orthogonal to stdin**, left
out deliberately rather than folded into a feature PR:

1. **`gori run intercept edit --raw 'X' --raw-file f.txt` forwards the FILE and never mentions
   `--raw`** (`cli/run/intercept.cr:465`). The same `if`/`elsif` shape this PR fixes for
   `repeater create`, in a different command. Verified by reading the chain.

2. **`repeater create --flow N` decides `is_ws` from the *flow's* request head, not the
   request being stored** (`cli/run/repeater.cr`, the `WsEngine.replayable?(detail.request_head)`
   call). With `--flow <ws-handshake>` plus a hand-authored request, the session gets the
   capture's outbound WS frames while `repeater send` decides from the stored request and never
   reads them. Reproduces in principle for `--request-raw` today, so it predates this PR — but
   I could **not** demonstrate the harm: HAR import carries no outbound WS frames, so the
   `is_ws && !ws_messages.empty?` guard kept anything from being written. Needs a real
   WebSocket capture to confirm before it is worth fixing.

Happy to file either as its own issue.

PR #1014, opened 2026-09-10
feat(decoder): read Java / .NET ViewState / PHP / pickle serialized blobs

### LINKED ISSUE #1011: feat(decoder): read native-serialization blobs — Java / .NET ViewState / PHP / Python pickle
## Summary

gori **detects** native-serialization blobs but cannot **read** them. The passive rule `src/gori/probe/passive/serialized_object.cr` flags a Java / .NET ViewState / PHP blob in client-controllable data and correctly warns about deserialization RCE — but the operator then stares at an opaque base64 string. There is no decoder that unpacks it into a readable structure.

The gap is exactly the shape the decoder catalog is built for. `Gori::Decoder` already renders MessagePack and CBOR as a JSON projection (`src/gori/binary_document.cr`, surfaced via `Pretty.try_binary_doc` in `src/gori/pretty.cr`). Native serialization formats are the same kind of thing — a serialized object graph — and want the same treatment: decode to a labeled tree, styled as JSON, with the raw bytes still the truth.

## Motivation

- Deserialization remains a top-severity, still-live enterprise bug class in 2026 (SharePoint ToolShell / .NET ViewState; recurring Java gadget-chain findings). When gori flags one, the very next question is "what's IN it?" — and today gori has no answer.
- Detection without a reader is a half-tool: the operator drops the blob into a separate CLI (`ysoserial`, a pickle disassembler, a ViewState decoder) to see field names, then comes back.
- `binary_document.cr` is the proven template: a reader that makes a JSON-shaped projection of a serialized graph, with types JSON can't hold coming back NAMED rather than folded away.

## Design

New codecs in the decoder catalog (`src/gori/decoder/catalog.cr`), each a **reader → labeled tree** projection, following the `BinaryDocument` model. Decode/inspect only — this issue does NOT build gadget chains (that stays an out-of-tool `ysoserial` step; gori's job is to show the operator what they're holding).

### Formats, in demand order

| Format | Wire marker | Projection |
|---|---|---|
| **.NET ViewState** (LosFormatter) | base64 of `FF 01 …` → `/wE…` (already the `VIEWSTATE` regex in the passive rule) | the ObjectStateFormatter triple/token tree; flag MAC-present vs absent |
| **Java serialized** | `AC ED 00 05` (`rO0AB…` base64) | stream header + class descriptors + field values as a named tree |
| **Python pickle** | opcode stream | opcode **disassembly** (à la `pickletools.dis`) — never execute; show `GLOBAL`/`REDUCE` calls, which is where the danger reads |
| **PHP serialize** | `O:len:"Class":n:{…}` / `a:` / `s:` | the typed structure PHP `serialize()` produces |

### Reader contract (mirrors `BinaryDocument.render`)

- Returns nil when the bytes are not the format their marker claims — that nil is the whole point (a PNG mislabeled as Java serialized must fall through to the hex view, not render as a plausible tree; this is the exact lesson in `pretty.cr`'s `try_binary_doc` comment).
- Never raises into the render path.
- Output is a JSON projection, styled `:json` (the `kind` override GraphQL/JWT/binary-doc already use).
- Bounded output (`MAX_OUT`-style ceiling) — a serialized bomb can't blow up the detail view.

### Surfaces (free, once the codec exists)

- **Decoder library / chain** — `java-deserialize`, `dotnet-viewstate`, `php-unserialize`, `pickle-disasm` become chainable converters (e.g. `base64-decode > java-deserialize`).
- **Pretty / detail view** — a body or field whose bytes carry a serialization marker gets the projection pane, the same way MessagePack/CBOR do now.
- The passive `serialized_object` finding can then point at "open in decoder" instead of just naming the format.

### P7

Raw bytes stay the truth and stay viewable; the tree is a lens. Pickle is disassembled, never executed. An encrypted/opaque ViewState (base64 that does NOT begin `/wE`) is deliberately left alone — the passive rule already declines it (`serialized_object.cr:22`), and the decoder does too.

## Where to implement

| Piece | File |
|---|---|
| Codec registration | `src/gori/decoder/catalog.cr` |
| Reader projections | `src/gori/decoder/serialized/{java,dotnet_viewstate,php,pickle}.cr` (new) |
| Projection model to mirror | `src/gori/binary_document.cr` |
| Pretty/detail wiring | `src/gori/pretty.cr` (`try_binary_doc` sibling) |
| Detection (existing) | `src/gori/probe/passive/serialized_object.cr` |

## Out of scope

- Gadget-chain **generation** (ysoserial-style payload crafting) — reading the blob is the gap; crafting stays external.
- Re-encoding an edited object graph back to the wire (Java/.NET especially are not worth a faithful serializer here). Read-only projection.

hahwul:
Closes #1011.

gori **detected** native-serialization blobs and could not **read** them: the passive
`serialized_object` rule points at a Java / ViewState / PHP blob in a cookie, a parameter or a
hidden field, and the operator was then holding an opaque base64 string. Four readers now
answer "what is in it?", with the projection `BinaryDocument` already defines.

## What landed

`src/gori/decoder/serialized.cr` — a shared `Reader` base (byte cursor, depth/item/output
ceilings, `bail`/`halt`, scratch renders) plus `Serialized.sniff`, the marker dispatcher.
Four readers beside it:

| Converter | Reads | Projection |
|---|---|---|
| `java-deserialize` | `AC ED 00 05` | `{"$object": …}` per instance, declared fields flattened super-first, the `writeObject` annotation, `$Proxy(…)` for a dynamic proxy |
| `dotnet-viewstate` | `FF 01` (`/wE…`) | the `ObjectStateFormatter` token tree, and whether a MAC is present |
| `php-unserialize` | `O:`/`a:`/… | `{"$class": …}`, private/protected names demangled with visibility kept beside them |
| `pickle-disasm` | opcode stream | `pickletools`-style disassembly, plus every callable named under `globals` |

Surfaces: the Decoder catalog (chainable — `base64-decode > java-deserialize`), and
`Pretty#try_serialized`, which renders a body carrying one of the four markers as the JSON
projection instead of a hex dump.

## Decisions worth arguing with

**A marker sniff, where MessagePack/CBOR deliberately refuse one.** `BinaryDocument.render`
dispatches on the content type alone because a wrong rendering costs more than a hex dump. None
of these four formats *has* a content type — a Java stream arrives as
`application/octet-stream`, a serialized PHP value as `text/plain` — so the bytes carry the
decision. Every marker is structural, and `describes?` still has to pass. The guide's "never a
sniff" paragraph is amended to scope itself to those two formats.

**Protocol-0 pickle is sniffed differently from how it is converted.** It has no header and its
opcodes are printable ASCII, so prose disassembles into plausible garbage: the sniff needs the
`\x80` PROTO opener, the converter does not. Same split `Codecs#document` already makes between
the content-type panes and a converter the operator typed.

**A ViewState's MAC is consumed and named.** There is no in-band marker for it —
`GetDecodedData` just strips the hash off the end — so a whole value tree plus a short trailing
run is a *signed* ViewState. Left as trailing bytes, `describes?` would (correctly, by its own
rule) refuse the majority of real ones and the pane would be dead for exactly the common case.
`"mac": false` is a finding, not the absence of one.

**Back-references are emitted, never expanded.** Java `TC_REFERENCE`, PHP `r:`/`R:`, pickle's
memo all come back as `{"$ref": n}`. Expanding one is both the output blow-up and the *less*
faithful rendering — the sharing is a fact about the graph.

**Read-only.** No re-encoding. Gadget-chain *generation* stays an out-of-tool `ysoserial` step.
An encrypted ViewState (base64 not starting `/wE`) is left alone, as the passive rule leaves it.

## Oracles, because a parser spec that agrees with its author proves nothing

- **pickle**: `pickletools.genops` diffed opcode-for-opcode and argument-for-argument across
  protocols 0, 2, 4 and 5 over 11 generated pickles — 100% agreement on the opcode stream. The
  remaining argument differences are the deliberate projection ones (a `GLOBAL` pair as a JSON
  array, a wide `LONG1` as `{"$bignum": …}` rather than a lossy number).
- **Java**: `javaobj-py3` parses every hand-built vector and agrees on handle numbering
  (`0x7e0003` for the object behind a two-field descriptor, etc.). It fails on two of them —
  a dynamic proxy, and `Externalizable` without `SC_BLOCK_DATA` — where gori's reader is either
  ahead of it or stops and says so.
- **.NET**: the token table is transcribed from `microsoft/referencesource`
  `System.Web/UI/ObjectStateFormatter.cs`, **fetched rather than recalled** — recall had
  `Token_IndexedStringAdd`/`Token_IndexedString` inverted and the `TypeRefAdd/AddLocal/Ref`
  triple missing entirely. The classic `/wEPDwUKLTE2MjY5MTY1NQ…` sample confirms it end to end.
- **PHP**: hand vectors with lengths computed rather than counted, including the `\0Class\0prop`
  and `\0*\0prop` manglings and a payload holding `"` and `;`.

## Three defects the crafted-repetition spec tier caught

The existing `binary_document_readers_spec` sweep (every byte, every prefix, 3 000 seeded
buffers) is the floor and structurally could not reach any of these: every buffer it makes is
**short**, and an output blow-up needs a small input **repeated**.

1. `Java#content` routed `TC_BLOCKDATA` past `step?`, so `77 00` repeated amplified ~13× with
   nothing charging the ceiling.
2. `Token_EventValidationStore` multiplied its count by 16 *before* checking it. Crystal's
   checked multiply raises `OverflowError`, which `Serialized.build` does not rescue and
   `CLI.run` does not either — a backtrace at the operator from four bytes of input.
3. `Php#unescape` called `Bytes.new(n)` on a declared length: `S:2000000000:"` is a 2 GB
   allocation in fifteen bytes, and the comment above it claimed the loop already prevented that.

Each is now pinned by a mutation-verified example. Reverting any one of the five guards fails a
spec: the two indexed-string tokens swapped (3 failures), the `TypeRef` triple swapped, the
version marker changed, `step?` dropped from `content`, the `n * 16` guard removed
(`OverflowError`), the `Bytes.new` guard removed (`total_bytes` moved 2 000 003 088 — the
rendering alone was identical either way, which is why the assertion is on allocation).

## Left out, deliberately

- **Field-level ViewState extraction.** `pretty.cr` formats *bodies*, and a ViewState lives in
  an `<input value=…>`, so the body sniff will rarely fire on real traffic. The Decoder tab is
  where an operator reads one today; wiring `decoded_view.cr` to offer the projection per-field
  is a separate surface and is not in this PR.
- **MS-NRBF.** A ViewState's `Token_BinarySerialized` payload is a `BinaryFormatter` graph — the
  ysoserial.net / ToolShell sink, and a whole format again. Its bytes come back named with the
  record header identified, rather than half-read.
- **Gadget-chain generation and re-encoding**, per the issue's own scope.

## Checks

`just check` (format gate clean; ameba full run unchanged on touched files), `just lint-gate`
(no changed file gained findings), `just test` (13 474 examples, 0 failures),
`just benchmark-check` (57 harnesses), `shards build`. `git log --all --not HEAD` was swept
first: none of the ~52 unmerged commits touches a serialization decoder.


hahwul:
Review pass applied — six correctness findings and five cleanups, each mutation-verified
(revert the guard, the spec fails).

**Correctness**

- **Headless parity.** `DecodedView#emit_binary_documents` called `BinaryDocument.render`
  alone, so a Java/ViewState/PHP/pickle body decoded in the detail pane and stayed opaque in
  `gori run show --format json` and MCP `get_flow` — the exact split that method's own
  docstring exists to prevent. It now falls back to `Serialized.sniff`, bounded by
  `MAX_SNIFF = Pretty::MAX_PRETTY` so the two surfaces agree about *which* bodies get a
  projection (the content-type readers run only on a body that asked for them; a marker sniff
  is offered every body there is).
- **ViewState indexed-string table desync.** A `Token_IndexedStringAdd` that is not UTF-8 was
  dropped from a POSITIONAL table, shifting every later `Token_IndexedString` — the index then
  resolved to a real but *wrong* string. Now pushed unconditionally with the base64
  substitution `read_type` already used next door.
- **Pickle counted its own gate byte as content.** `got!` fired on `PROTO`, the byte
  `Serialized.sniff` gates on, making `describes?`'s third test vacuous for this format: a
  five-byte `PROTO` + truncated `FRAME` replaced the hex view with a disassembly of nothing.
  `PROTO`/`FRAME` are the stream header, so they no longer count, and neither does an opcode
  whose argument ran out. The spec that pinned the old accept is flipped.
- **PHP `S:` was not treated as a string key.** The hex-escaped form is the standard spelling
  used to get past a filter looking for `s:`, so the crafted input this reader exists for was
  the one rendering wrong — the key's whole JSON text, quotes included, became the member name.
- **Negative `Token_TypeRef`.** `encoded_int` sign-extends and Crystal's `Array#[]?` counts a
  negative index from the END, so `-1` resolved to the most recent type: a name the stream
  never wrote, shown as if it had.
- **A fourth uncharged loop.** `event_validation`'s hash loop emitted without `step?`; a
  250 000-entry store rendered 8.75 MB against a documented 8 MiB ceiling. Now in the bomb tier
  beside the three found before it.

**Also**

`Reader#scratch` swaps `@sink` under an `ensure` (leaving it on a discarded buffer loses the
*ceiling*, not a rendering); the dead `@root` ivar is gone; `binary_formatter?` uses a slice
compare; the module doc, both guides and the CHANGELOG no longer claim pickle's memo comes back
as `$ref` (it is a disassembler — a `BINGET` is an opcode record); `field_descs` validates the
type code before reading the name, so a rewind lands on the offending byte rather than one byte
inside a field name; `e7` lives once in `spec/support/serialized_vectors.cr`; `DOC_READERS` is
`private`; the perf example's calibration comment matches the six readers it now runs.

Not changed: the `GC.stats.total_bytes` oracle in the PHP allocation spec is process-global and
therefore order-dependent, which is a real caveat — kept because the rendering is byte-identical
with the guard removed (only the allocation differs) and the signal is 2 000 003 088 bytes
against 16 MiB of headroom. The spec says so.

`just test` 13 480 examples / 0 failures, `just check` format gate clean, `just lint-gate` no
file gained findings, `just benchmark-check` 57 harnesses, `shards build`.

PR #1015, opened 2026-09-10
feat(jwt): sign/verify RS/PS/ES/EdDSA, and recognize encrypted JWE (#1010)

### LINKED ISSUE #1010: feat(jwt): sign/verify ES/EdDSA/PS algorithms and decode JWE (encrypted JWT)
## Summary

gori's JWT tooling (`src/gori/jwt.cr`, `src/gori/jwt/{forge,attacks,present}.cr`; MCP `jwt_decode` / `jwt_encode` / `jwt_attacks`; `gori run jwt`) covers HMAC well but stops short of what real OIDC deployments now ship:

- **Signing/forging is HMAC-only.** `Jwt::Forge::ALGS = %w[HS256 HS384 HS512 none]` (`src/gori/jwt/forge.cr:21`). An operator cannot forge or re-sign an **`ES256` / `EdDSA` / `PS256`** token even when they hold the key — and ES256 is the default for a large share of modern IdPs.
- **JWE (encrypted JWT) is absent entirely.** A five-part `header.encrypted_key.iv.ciphertext.tag` token is not recognized as a JWT at all: `JWT_RE` in `src/gori/pretty.cr` matches only the three-part JWS shape, so a JWE renders as an opaque dotted string with no pane, no header decode, nothing.

This is the natural completion of tooling that already decodes, forges, and attacks JWS.

## Motivation

- ES256/EdDSA are the common case in current OIDC (Apple, many Auth0/Okta/Keycloak configs). "gori can read your JWT but can't sign the one your test needs" is a sharp edge on an otherwise complete tool.
- JWE tokens are increasingly used for ID tokens and userinfo. Today they are invisible — the operator cannot even see the `alg`/`enc` header to know what they are looking at, which is the first thing they need.
- The seam is clean: `present.cr` is the single JSON source of truth for CLI + MCP (the DecodedView lesson), so extending the primitive lights up every surface at once.

## Design

Two independent pieces; ship in this order.

### 1. Signature algorithms beyond HMAC (JWS)

- **Verify + re-sign** for `ES256` / `ES384` / `ES512` (ECDSA), `PS256` / `PS384` / `PS512` (RSA-PSS), and `EdDSA` (Ed25519), using OpenSSL primitives already available in Crystal's stdlib (`OpenSSL::PKey` / `Digest`).
- `Jwt.sign` and `Forge::ALGS` grow to cover them; `forge.cr`'s `HMAC_DIGEST` map is joined by an asymmetric path keyed on `alg`.
- Key input: PEM private key (file path or inline) for signing; PEM public key (or a JWK / a JWKS URL — outbound, so through `Gori::Outbound`'s scope gate, operator-initiated) for verify.
- `jwt_attacks` gains the **algorithm-confusion** family properly: RS/ES/PS public key → HS secret (already stubbed toward in `attacks.cr:174` with a `kty:RSA` JWK), which needs the real public key to be meaningful.

### 2. JWE (encrypted JWT) — decode first, forge later

- **Recognize** the five-part compact JWE shape and surface the protected header (`alg`, `enc`, `kid`, `epk`) — this alone is most of the value: the operator learns what the token is.
- **Decrypt** when the operator supplies the key: content-encryption `A128GCM` / `A256GCM` / `A128CBC-HS256` etc., key-management `dir`, `RSA-OAEP`, `ECDH-ES` (the common set). Missing key → header shown, body marked "encrypted (no key)", never a guess.
- **Encrypt/forge** is the last step and can be a follow-up — decode/inspect is the gap that bites first.
- Wiring: `JWT_RE` / the JWT sniff in `src/gori/pretty.cr` learns the five-part shape; a new `src/gori/jwt/jwe.cr` holds the crypto; `present.cr` gains a `jwe` projection shape parallel to `decode_json`.

### P7 / safety

- Raw token bytes remain the truth. A JWE with no key shows its header and stops — it never fabricates a plaintext.
- Asymmetric signing requires an operator-supplied private key by construction; gori generates nothing on its own.

## Where to implement

| Piece | File |
|---|---|
| Sign/verify primitives (EC/PS/Ed) | `src/gori/jwt.cr`, `src/gori/jwt/forge.cr` |
| Attack families (real alg-confusion) | `src/gori/jwt/attacks.cr` |
| JWE crypto | `src/gori/jwt/jwe.cr` (new) |
| JWT/JWE sniff | `src/gori/pretty.cr` (`JWT_RE`, `try_jwt`) |
| JSON projections (CLI + MCP) | `src/gori/jwt/present.cr` |
| CLI / MCP surfaces | `gori run jwt`, MCP `jwt_*` tools |

## Out of scope

- A key-management UI. Keys are supplied per-invocation (path or inline), like every other operator-supplied secret.

hahwul:
Closes #1010.

## What ships

**1. Signature algorithms beyond HMAC.** `RS256/384/512`, `PS256/384/512`, `ES256/384/512` and `EdDSA` sign *and* verify, with a key the operator supplies — inline PEM or a path to one. gori generates nothing: forging an `RS256` token still requires the `RS256` private key. `Jwt::ALGS` grows from 4 to 14, which lights up the TUI alg cycle, the MCP `alg` enum and the CLI in one move.

`gori run jwt --verify` / MCP `jwt_verify` is the new surface: *would a server holding this key accept this token*. It verifies under the alg the **token** declares, never one the caller picks, because that is what a vulnerable server does. Verification accepts a public key, an X.509 **certificate** (what an IdP publishes as `x5c`) or the private key.

With the server's public key, `--attacks --key` / `jwt_attacks{public_key}` adds the **algorithm-confusion** family: `HS256` re-signs keyed with the public key's own bytes, one payload per spelling a server might hold — the canonical SPKI PEM OpenSSL would write (which is also how a supplied certificate is reduced to its public half) and the bytes as given, with and without a trailing newline.

**2. JWE, recognize-only.** A five-part `header.encrypted_key.iv.ciphertext.tag` token used to render as an opaque dotted string at every surface. It is now recognized wherever a JWS is (JWT tab, Decoder, body pretty-printer, `gori run jwt`, MCP `jwt_decode`) and shows its protected header — `alg`, `enc`, `kid`. gori does not decrypt; the claims are marked encrypted and nothing is guessed at in their place.

## Deliberately not in scope

Called out because the issue lists them:

- **JWE decryption and encryption.** `dir` / `RSA-OAEP` / `ECDH-ES` key management plus `A*GCM` / `A*CBC-HS*` content encryption is a second FFI subsystem (AEAD tag ctrls, `EVP_PKEY_derive`, the CBC-HMAC composite). The issue itself stages decode first — "decode/inspect is the gap that bites first" — and that is what this is.
- **JWK and JWKS-URL key input.** A JWKS URL is an outbound fetch and would drag `Outbound` and the scope gate into a pure-compute command for a parameter the issue lists parenthetically. Building an `EVP_PKEY` from raw JWK parameters is its own pile of FFI. PEM and X.509 cover the operator who holds a key.
- **Ed448.** No JOSE `alg` is registered for it; the key-type check names it rather than pretending.

## Shape notes

**The issue's premise about the stdlib is wrong.** Crystal 1.21 has no `OpenSSL::PKey` — the stdlib binds no `EVP_PKEY` operations at all. Everything asymmetric is FFI in a new `src/gori/jwt/asym.cr`, following the precedent in `src/gori/oast/rsa.cr`. Three things that compile fine and fail at runtime, each verified against real keys before the code was written:

- **ECDSA signature format.** OpenSSL emits DER `SEQUENCE{r,s}`; JOSE wants fixed-width `r‖s`. `ES512` is 132 bytes, not 128 (P-521 is 66 per component). Converted both ways, with a DER-integer leading-zero case the spec pins directly.
- **Ed25519 is one-shot only** and takes a NULL digest. The `EVP_DigestSign`/`EVP_DigestVerify` one-shot pair is used for every algorithm — it is also the one spelling that links on both 1.1.1 (where `EVP_DigestSignUpdate` is a macro) and 3.x.
- **PS\* needs two ctrls on the pctx** or OpenSSL signs PKCS#1 v1.5 *silently* — a valid RS256 signature under a PS256 header, which nothing accepts and nothing reports. `EVP_PKEY_CTX_ctrl` with keytype/optype `-1`, matching `oast/rsa.cr`; the numeric constants come from `rsa.h` on the box, not from recall.

`EVP_PKEY_base_id` / `EVP_PKEY_bits` were renamed to `EVP_PKEY_get_*` in OpenSSL 3.0 with the old names left as macros, which do not link. Guarded on `compare_versions(OPENSSL_VERSION, "3.0.0")`.

**FFI hoist.** `lib LibCrypto` is reopened in three files now, and a `fun` declared twice collides. The generic memory-BIO and `EVP_PKEY_CTX` funs moved from `oast/rsa.cr` up into `proxy/tls/ffi.cr`, which its header now states is the shared home. OAST's RSA path is covered end-to-end by `spec/oast/engine_spec.cr` (keygen, SPKI export, OAEP decrypt of a fixture ciphertext), so the move is runtime-verified, not just link-verified.

**The derived projections were the real work.** Widening "this is a JWT" without sweeping every projection built on it is how a JWE ends up read as a JWS with its *wrapped key* in the claims slot. Each of these refuses or branches now, and each has a spec that fails when the guard is removed:

- `Jwt.attacks` returns nothing for a JWE — every gate would otherwise pass and then splice the wrapped key in as `payload_seg`.
- `decode_json` gains a `type` discriminator: `"JWS"` keeps the old fields, `"JWE"` reports `payload: null` with `encrypted: true` rather than a null that reads as "unparseable". A JWE previously would have emitted `signed: true` with the IV segment as `signature`; that pair is gone for JWEs.
- `Jwt.verify` says a JWE carries an AEAD tag, not a signature, instead of "unsupported alg RSA-OAEP".
- `Decoder::Codecs.jwt_decode`'s existing "5 parts, JWE-shaped" *warning* becomes the JWE branch; a five-part blob whose header declares no `enc` still gets the warning.
- **A pre-existing false positive is fixed**: `probe/passive/jwt.cr` read a JWE's key-management `alg` (`RSA-OAEP`, `dir`, `ECDH-ES`) as a signing alg and reported every encrypted token as `jwt_weak_alg`, "signed with a non-standard algorithm". `enc` is the discriminator (RFC 7516 §4.1.2 requires it; no JWS defines it), which also works on the truncated three-segment prefix the scan yields.

`Jwt::SCAN_RE` now runs to five segments so a JWE matches whole rather than being cut after its IV. That greed can swallow the word after a token ending a sentence (`…a.b.c.Next`), so `Jwt.narrow` steps a non-JOSE match back down to the JWS segment counts before dropping it — a token that used to be found still is.

**`HMAC_DIGEST` stays symmetric-only, on purpose.** `attacks.cr` gates its "SECRET FOUND" claim on `HMAC_DIGEST.has_key?`; an asymmetric alg in that map would turn an HMAC coincidence over an RSA signature into a claim that gori had recovered the server's key. A spec pins the map's exact keys.

**Key material is never echoed.** An operator who passes an HMAC secret to an asymmetric alg gets "key is neither an inline PEM block nor a readable file path" — never their own bytes back, in the message or in whatever log captured it.

## A bug found in review, fixed in the second commit

`sign` reaches `HMAC_DIGEST` before it reaches `Asym`, so `--alg HS256 --key ./server.pub` HMAC-signed the **path string** and reported a token signed with a filename, with nothing to say otherwise. `Jwt.key_material` now resolves the `--key`/`key` spelling to its PEM at both surfaces before the engine sees it — which also makes that spelling mean what it looks like: an algorithm-confusion token keyed with the public key's own bytes. `--secret` stays a literal; a `--key` that is neither PEM nor a readable file is refused rather than signed with.

## Verification

- **Ed25519 is the anchor.** It is deterministic, so RFC 8037 Appendix A.4 is a byte-exact known-answer test — the one assertion here that catches a digest-sign path that is wrong but self-consistent. ECDSA and RSA-PSS are randomized, so they get round-trips, tamper-rejection, and a width assertion run 20× to catch the short-`r` case. PS-vs-RS cross-verification over one key is the only local way to prove the PSS ctrls took effect.
- Key PEMs are **committed fixtures** (`spec/support/jose_keys.cr`), not generated at spec time: a fresh key per run makes a randomized-signature failure unreproducible.
- Every source change was **mutation-tested**: 20 mutations across the crypto path, the JWE sweep, the projections, the CLI/MCP glue and the two TUI panes; each produced failures, each was reverted. A spec that passes with the code broken is not a gate.
- The TUI was driven headlessly (tmux) end to end: the KEY card, a PEM path typed into it, live RS256 signing, and a JWE decoding in the DECODE lens with ATTACKS correctly at 0.
- `crystal spec` 13,498 examples green; `crystal tool format --check src spec bench scripts`, `scripts/ameba_gate.sh`, `scripts/bench_check.sh`, `scripts/nix_shards_check.cr` and `shards build` all green.

Docs: the JWT guide and the attack-a-JWT playbook gain an Algorithms table and a JWE section, the CLI reference gains `--verify` / `--key`, and the MCP tool table gains `jwt_verify` — English and Korean both.


hahwul:
## Review round

A `/code-review` pass over the branch raised 15 findings. Each was reproduced against the built binary or a driver script before anything was changed, then the fix was mutation-tested. Nine were real and are fixed in bc18a3ac; four were correct observations I acted on differently; two are declined with reasons.

### The one that mattered

**The passive rule's JWE gate was `header.has_key?("enc")` alone — and `enc` is a header parameter anyone can add.** One field on an otherwise ordinary three-segment token suppressed every check in `probe/passive/jwt.cr`, including the **High** `jwt_alg_none`:

```
alg:none, no enc  -> ["jwt_alg_none", "jwt_no_expiry"]
alg:none + enc    -> []
```

That is detection evasion, and this PR introduced it. The gate is now the full predicate — five segments **and** `enc` — with the dot count short-circuiting before the parse so an ordinary token pays nothing. My original comment justified the loose form with "it also works on the truncated three-segment prefix `SCAN_RE` yields", which was stale reasoning: the same PR widened `SCAN_RE` to five segments, so the precise predicate was available and I did not go back and use it.

### Also fixed

- **`verify` vouched for a 4-segment token.** `header.payload.sig.SMUGGLED` returned `verified: true` — the HMAC over parts[0..1] matched parts[2] and the fourth segment was dropped on the floor. `Codecs.jwt_decode` and `attacks` both refuse or surface extra segments; `verify` was the outlier.
- **The `verified:` line printed the token's own `alg` unescaped.** `alg` is captured, attacker-chosen text; a header of `{"alg":"<ESC>[2J<ESC>]0;pwn<BEL>"}` cleared the operator's screen and rewrote the window title. The `reason:` line beside it was already sanitized — this was the one line that was not. The text form is now a pure `jwt_verify_lines(v)` so it is assertable rather than source-grepped.
- **A NUL byte in a key path escaped as `ArgumentError`.** `File.file?` raises it, it is not an `IO::Error`, so it went past every `rescue ForgeError` — reaching MCP's blanket rescue as `INTERNAL` ("gori is broken") for a mistake in the agent's own argument, and crashing the TUI's JWT tab.
- **`JWE_RE` had the ciphertext and tag quantifiers inverted.** RFC 7516: the AEAD tag is always present; the ciphertext is empty when the plaintext is. A JWE over an empty payload therefore fell out of the shape and was then reported as "the header declares no `enc`" — which its header plainly did. The message no longer asserts a reason nothing checked.
- **A bad `public_key` raised for an RS256 token and was swallowed for an HS256 one**, because the alg gate ran before the key load. MCP, which has no other key resolution, got the silent half. The docstring's promise is that the operator sees their typo; now it holds either way.
- **The TUI's single key field was silently reinterpreted across the HMAC/asymmetric boundary.** That one card holds two different things — a literal HMAC secret, or a PEM key gori resolves — and which one comes from the alg alone. `Ctrl-A`, or a `Load decoded claims` that adopts the token's alg (newly possible now that `ALGS` contains RS256), carried a typed `./private.pem` across into HS256, where it became a fourteen-byte HMAC secret and OUTPUT showed a token signed with a filename. The field is now dropped on a boundary crossing, and the status line says so. This is the same class as 9ee07791; the CLI avoids it by having two flags, and the TUI has one field, so the boundary is where its content stops being meaningful.
- **The three-reader PEM fallback left OpenSSL's error queue dirty.** On the success path one or two readers fail and push entries; gori never reads them, but Crystal's `OpenSSL::Error` does, so a stale PEM parse error could surface later as the reason a proxy TLS handshake failed. Cleared on every BIO teardown.
- **The ATTACKS empty-state hint parsed the token on the render path** — every frame, for the whole time a token is being typed, which is exactly when that pane is empty. It is cached beside `attacks` now, computed once per edit.
- **`forge.cr` and `attacks.cr` referenced `Jwt::Jwe` without requiring it**, compiling only because an unrelated file happened to pull it in. Latent, but the bench tree has been bitten by this class before.

### Acted on differently

- **`--decode --key typo.pem` aborted over a flag that mode ignores**, and `--attacks --key` opened the key file three times. Rather than special-case, `key_material` moved into the branches that use a key, and `attacks` is handed the already-resolved PEM: one read, and no refusal from a mode with no use for one.
- **The EC key check is by bit width, so a secp256k1 key passes as ES256.** Real, and I could not fix it portably — `EVP_PKEY_get_group_name` is 3.0-only, and the 1.1.1 route to the curve NID goes through EC_KEY accessors 3.0 deprecates. What I did fix is the message, which claimed a NIST curve it had never checked for: a brainpoolP384r1 key was reported as "P-384". It now reports the width it measured, and the code comment names the gap.
- **The MCP `key` / `public_key` path argument is a file existence and size oracle for a `--read-only` client.** I kept the path form — the issue specifies it, and `cookie_crack`'s `wordlist` is the same shape — but the docs claimed these tools "touch no network or state", which is no longer accurate. English and Korean now say they touch no network and write nothing, and that a key given as a path is read from disk.
- **`recompute_encode` re-reads and re-parses the PEM on every keystroke** under an asymmetric alg, allocating an `EVP_PKEY` per call. Real, and I left it: ~1ms of RSA per keypress on a human-paced field is not a stall, the objects are GC-collected rather than leaked, and an mtime-keyed cache is invalidation state that can go wrong in its own way. Worth revisiting if it is ever felt.

### Declined

- **`--attacks` on a JWE exits non-zero.** The suggestion was to make it an empty list at exit 0, on the grounds that `--verify` answers rather than fails. But `--attacks` already aborts for *any* token it cannot derive payloads from; special-casing JWE to exit 0 would leave a non-JWT still aborting, which trades one inconsistency for another. The message names the reason.

### Verification

Every fix was reproduced first and mutation-tested after: 11 mutations across the probe gate, `verify`, the escape handling, the NUL guard, the JWE shape, the key-load ordering, the TUI boundary clear and the cached render flag — all killed, all reverted.

One mutation **survived**, and that was the point of running them: disabling `verify`'s new segment guard changed nothing, because an earlier partial edit of mine had left the guard in the file **twice**. The loop found duplicated code that both the suite and a reading had passed over. Removed, re-mutated, killed.

`crystal spec` 13,509 examples green; format-check, ameba diff gate, bench-check, nix-shards and `shards build` all green.

PR #1134, opened 2026-09-21
fix(tui): stop a wide glyph shearing the screen (#1125)

### LINKED ISSUE #1125: Crashes on proot Ubuntu in Termux, and also on Parrot 7.3 echo
https://x.com/Xc12vm/status/2101908667268276519

hahwul:
Closes #1125.

The reporter called it a crash, but the screenshot is a live TUI with a sheared screen: every row's left ~21 columns hold the previous row's tail, and the body sits 21 columns right.

Two termisu bugs stack up.

- `advance_cursor` works out the next position by wrapping at termisu's own width, and `move_cursor` skips the escape when that position already matches. A 40×6 diff frame goes out as `\e[1;1H` plus 240 contiguous characters, so one glyph the terminal measures differently pushes every row below it over. `sync` just redraws the glyph and shears again.
- `wide_codepoint?` never looks at the BMP symbol area. Measured against tmux, 155 codepoints came back a column narrower than the terminal draws them, 151 of them `EastAsianWidth=W`. `✅ ⭐ ❌ ⚡ ❓ ❗` are all in that set.

Repro: an issue titled `✅⭐❌⚡☕⏳✨➕❓❗ width probe`, Issues tab in tmux 87x30, then resize to 86. Before, everything below that row sat 12 columns over; after, aligned.

It never showed up locally because the mismatch set depends on the terminal's own width table, and gori's chrome glyphs all agree at 1 column. It takes captured data, which matches "after a few requests come in".

### The dependency move

The lock goes df6e907 (0.6.1) to b790d91 (0.6.3), which is **34 commits**, not just the two fixes. `shard.yml` pins termisu by `branch: main`, so the lock is the only pin. That range rewrites `input/parser.cr` (+629), `terminal.cr` (+385), `termisu.cr` (+389), the logging backend and the resize source, and adds a process-global terminal lease (`TerminalInUseError`) and input-ownership enforcement.

The two fixes are on `fix/cursor-wrap-anchor` and `fix/east-asian-wide-table`, open as drafts upstream. What I audited inside the range: the input source and parser (both carried patches live there), the `Termisu.new` failure set, and the new width table. I did not re-audit the logging rewrite, terminfo/tparm or the mode-scope gate.

### Both carried patches had drifted

The pin guards exist so a bump forces a re-read, and this time both were right to fire.

- **`input_idle_backoff_patch.cr` had stopped applying.** Upstream `run_loop` grew two parameters, so the carried no-argument copy became an overload nobody calls and the back-off was gone with no compile error. Its frozen body also predated `@pending_event`. Rebased, and its "what this widens" note rewritten: the double-fiber race it disclosed is fixed upstream, and the real new cost is that `stop` now waits on the sleep, so a mode transition blocks 16ms instead of 4ms.
- **`paste_end_marker_patch.cr` froze a method upstream has since fixed.** 51cc502 rewrote `read_paste_end_tail` around `@reader.read_byte(paste_wait_ms)`, which is the exact zero-budget bug the patch header described. The carried copy still had the old `wait_for_data` shape, so the next upstream fix inside it would have been reverted silently. Override dropped, header rewritten; `parse_paste_escape` and `leave_paste!` stay, because upstream still has no give-up branch and dropping them fails two behavioural examples.

Both specs now compare each override's signature against the installed shard, so the next drift fails instead of passing over a frozen copy. `open_terminal` also widened to `Termisu::Error`, since `TerminalInUseError` and `InputOwnershipError` arrived with this bump and are not `IO::Error`.

### Verification

TUI 4690, cli+mcp+rules 1953, other dirs 4139, all 0 failures. Root-level has 4 failures, all pre-existing on this machine. Format, bench, nix-shards and ameba gates pass.

Not here: the gori-side recovery path (`view.refresh` has no chord, and a sheared row stays wrong until its content changes), and a third termisu fix for the resize source that I could not demonstrate.


hahwul:
upstream PRs
- https://github.com/omarluq/termisu/pull/203
- https://github.com/omarluq/termisu/pull/204

